PR #10, opened 2016-10-26
Typo correction

Denino:
Correction that Android Studio won't show error: "Couldn't resolve resource @id/play_button"


PierfrancescoSoffritti:
thanks :) the fix has just been added to version 0.7.5

Denino:
Very good :) Thanks for great library!

PR #37, opened 2017-09-29
isPaused method

xitorrito:


PierfrancescoSoffritti:
mmm I'm not sure this is the best way to do it, I think a `getPlayerStatus `method would be much better. This method should get the status directly from the Javascript player.

Also the `YouTubeListener `interface already has a `onStateChange` method, this could be used to keep track of the status, from outside the player.

PR #38, opened 2017-10-06
cleanup Handler callbacks on destroy

talklittle:
Fixes a crash on some Android 4 devices when trying to execute JavaScript after the WebView has already been destroyed. E.g. when calling `pauseVideo()` followed by `release()`.


PierfrancescoSoffritti:
thanks a lot :)

PR #57, opened 2017-11-16
Add setVolume API

### LINKED ISSUE #13: No Volume Contol
Hello,
As i dig into more using AndroidYoutubePlayer, and i really like it, but i'm facing one issue, is there any controls available in player for controlling the volume of video programatically, like setVolume(), etc.

talklittle:
https://developers.google.com/youtube/iframe_api_reference?hl=en#setVolume

Fixes #13

### DIFF
--- a/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/player/WebViewYouTubePlayer.java
+++ b/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/player/WebViewYouTubePlayer.java
@@ -121,6 +121,20 @@ class WebViewYouTubePlayer extends WebView implements YouTubePlayer {
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


PierfrancescoSoffritti:
Hi, thanks for the pull requests.
As I said in issue 13, on Android the volume is controlled by the system, not by single apps. Therefore I don't think adding this method is a good idea.

talklittle:
That's reasonable. How about, at a minimum, we always set the volume to 100% before playing? It's not clear if YouTube does or will ever change the default volume.

PierfrancescoSoffritti:
Ok, I have thought about it and I've reconsidered. While single Activities are not supposed to control the volume, it's probably best to have that option, then it will be up to the users of the library to build a good user experience.

Thank you again for contributing! If you're building an app using this library let me know if you want it added to the list in the readme :)

PR #88, opened 2018-02-28
periodically updates of current playing time using setTimeout instead of setInterval

beyondeye:
Hi thanks a lot for this library!
While using it, I found that in some cases the timerTask was not correctly removed. I was getting a deluge of updates of the current playing time and the app performance was suffering. So I what I did? I substituted
it with setTimeout, and I have also added a method sendCurrentSeconds() to YouTubePlayer interface to ask the player the current time, even when player is paused.

### DIFF
--- a/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/player/WebViewYouTubePlayer.java
+++ b/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/player/WebViewYouTubePlayer.java
@@ -143,6 +143,15 @@ class WebViewYouTubePlayer extends WebView implements YouTubePlayer, YouTubePlay
             }
         });
     }
+    @Override
+    public void sendCurrentSeconds() {
+        mainThreadHandler.post(new Runnable() {
+            @Override
+            public void run() {
+                loadUrl("javascript:onCurrentSeconds()");
+            }
+        });
+    }
 
     @Override
     @PlayerConstants.PlayerState.State
--- a/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/player/YouTubePlayer.java
+++ b/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/player/YouTubePlayer.java
@@ -32,6 +32,11 @@ public interface YouTubePlayer {
      * @param time The absolute time in seconds to seek to
      */
     void seekTo(final int time);
+    /**
+     * trigger the {@link YouTubePlayerListener#onCurrentSecond(float)}  callback
+     * with the video current playing time in second
+     */
+    void sendCurrentSeconds();
 
     @PlayerConstants.PlayerState.State
     int getCurrentState();
--- a/YouTubePlayer/src/main/res/raw/youtube_player.html
+++ b/YouTubePlayer/src/main/res/raw/youtube_player.html
@@ -67,9 +67,6 @@
     	}
 
     	function sendPlayerStateChange(playerState) {
-            var timerTaskId;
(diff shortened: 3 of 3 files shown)


PierfrancescoSoffritti:
Hi! thank you for the pull requests!

Interesting, I wasn't aware of the issue with setTimeout.

With the changes you are suggesting the "currentTime" of the video is sent out automatically only once, after that the user of the library will be required to periodically call `sendCurrentSeconds()` in order to update the seekbar of the player, have I understood it correctly?

PierfrancescoSoffritti:
There is an easier solution to the problem. I've realized that the bug is caused by a mistake in the code.

This is the current implementation:

```
function sendPlayerStateChange(playerState) {
    var timerTaskId;
    clearTimeout(timerTaskId);
    ...
}
```

obviously declaring `var timerTaskId` before clearing the timeout makes no sense, the variable loses the value it had previously, therefore the timeout is never stopped.
The variable has to be declared only once outside of the method.

I think this change is the right one to make, instead of modifying how the library works.
Let me know if you want to be the one sending the pull request with that change, otherwise I can commit it myself :)

(If you sent a pull request, sent it to the dev branch please)

beyondeye:
The proposed changes do no change the library behavior: the updates will be sent every 100ms as before.
Every 100ms a new timeout is set.
Note that setInterval is problematic in itself, not because of the bug. See this link

(http://reallifejs.com/brainchunks/repeated-events-timeout-or-interval/)

PierfrancescoSoffritti:
I didn't notice your `setTimeout ` is recursive, my bad.
Next time try to separate PRs based on what they're changing. I thought the `setTimeout` and `sendCurrentSeconds()` changes were related, but they're not, it's a bit confusing if they come in the same PR.

- `setTimeout` change:
I'm aware of the different behavior the two functions have. I think in this case `setInterval` works well because virtually no work happens at each interval. They both have their [advantages and disadvantages](https://stackoverflow.com/questions/729921/settimeout-or-setinterval).
I believe the determinant factor here is readability, I find the `setInterval` solution to be more readable. What do you think?

- `sendCurrentSeconds()` addition:
Why do you think this method would be useful? The user can already keep track of the current time if he needs to.

beyondeye:
Sorry, I should have split the PR. You are right. Why the need of sendCurrentSeconds()? When the video is paused it is not possible to get the current second. If the user play with the youtube slider while the video is paused, for example, or after a seekto command. Now that I checked in my code, I saw that I am not using it any more, so if I want I can change the PR.

beyondeye:
By the way is Italian an allowed language for discussing PR? I am also Italian

PierfrancescoSoffritti:
Let's stick to English, so everyone can understand.

I think is best if the user of the library is the one responsible for keeping track of the current time, it can be done extremely easily. (ie. save it in an external variable every time the callback is called).
This way the interface of the player is kept clean. I've already made this mistake with the `getCurrentState` method, the user should be responsible of keeping track of the state (if he needs to), not the library.
Methods like `getCurrentState`, getCurrentTime, getVolume etc add unnecessary noise to the interface.

Also consider that if the progress bar is moved, or seekTo is called the current time of the video actually changes, therefore the "previous current time" is generally useless. For specific use cases in which it's not useless, the user can decide to save how many time stamps he wants.

beyondeye:
Fine by me. So do you want a modified PR without SendCurrentSeconds()? On the dev branch? By the way the dev branch look out of sync with respect to master (at least with respect of the demo app)

PierfrancescoSoffritti:
Yeah, you're right. Last week I've updated the sample app on the main branch. I guess at this point is easier if I just push everything myself.

The new version should be up in a few minutes! with the bug fixed! thanks for the help :)

PR #152, opened 2018-06-21
feat: Make Live mode UI and behaviour in sync with YouTube native app

rajeefmk:
The changes included are : 
1. Introduced Live-VOD (Video on Demand) mode similar which is enabled when user moves the seekbar during live stream. 
2. Live stream indicator tag is made clickable and now has the same functionality as that of native app wherein it'll restore the live state. 
3. Seekbar is always visible now and during Live stream, it's progress position value will be that of the video duration.
4. Position of  Player control UIs are changed to mimic that of the native app.

### DIFF
--- a/YouTubePlayer/build.gradle
+++ b/YouTubePlayer/build.gradle
@@ -14,6 +14,11 @@ android {
         versionName "7.0.1"
     }
     buildTypes {
+        debug {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
+
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
--- a/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/ui/DefaultPlayerUIController.java
+++ b/YouTubePlayer/src/main/java/com/pierfrancescosoffritti/youtubeplayer/ui/DefaultPlayerUIController.java
@@ -71,6 +71,10 @@ public class DefaultPlayerUIController implements PlayerUIController, YouTubePla
     private boolean showPlayPauseButton = true;
     private boolean showBufferingProgress = true;
 
+    private boolean isLive = false;
+    private boolean isLiveVOD = false;
+    private int videoDurationValue = 0;
+
     public DefaultPlayerUIController(@NonNull YouTubePlayerView youTubePlayerView, @NonNull YouTubePlayer youTubePlayer) {
         this.youTubePlayerView = youTubePlayerView;
         this.youTubePlayer = youTubePlayer;
@@ -108,6 +112,7 @@ public class DefaultPlayerUIController implements PlayerUIController, YouTubePla
         playPauseButton.setOnClickListener(this);
         menuButton.setOnClickListener(this);
         fullScreenButton.setOnClickListener(this);
+        liveVideoIndicator.setOnClickListener(this);
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


rajeefmk:
@PierfrancescoSoffritti Please consider this as a humble way of thanking you for this amazing library which helped me push out my app to playstore in a short duration. The player UI customization was a kickass feature and was really helpful. 

I wouldn't have known about the possibility of using iFrame API in native app if not for this wonderful library and would have gone with the player fragment view (which comes with all the native code dependencies).

I would also like to actively contribute to this library and help you resolve issues. Please feel free to assign me to any issues in case your plate gets full. 

Cheers,

PierfrancescoSoffritti:
Hi! I'm glad you've found the library to be helpful :) I'd love to see your app!

Thank you for the contribution!

I'm testing your branch, [this](https://dl2.pushbulletusercontent.com/j7HdY0HVJhvZsYPdKDISJytXyUJd50kZ/Screenshot_20180621-110259.png) is the live UI I'm seeing. Is this correct? I can't see any live indicators. Also i'm not sure current time and duration are displayed correctly.

I've seen that you changed the default UI of the player, this should go into a separate PR, as it's not related to live mode. Also I'm not sure moving the seekbar at the bottom is a good idea, it would require further discussion.

rajeefmk:
@PierfrancescoSoffritti Thanks for the feedback :)  I had made a last minute change related to resetting live status when resetUI() method was called, but it seems unnecessary. 

About the UI changes, unfortunately the functionality change is directly tied to the UI also and hence one can't be tested without the other. So decided to combine them into single PR. 

And I agree absolutely that said the UI changes needs to be discussed more. 

I'm Attaching the new live screenshot, live-VOD screenshot as well as screenshot from native app. 

Please do share your thoughts. 

live_stream
live_vod_stream
native_app

PierfrancescoSoffritti:
Nice, i'll try it again later!

Why is the position of the seekbar related to live mode?

PierfrancescoSoffritti:
I've reviewed the changes you've made. I like it, it's a very welcomed improvement over the old live UI! good job :)

A few notes/questions:
1. I'd like the XML file defining the UI of the player to not be changed. If i'm not mistaken the only UI addition needed to use your improved live UI is the red/gray dot in the LIVE textview, which is added as a drawable on the textview. Therefore no change should be required on the XML file, am I correct?
2. Why does the `PlayerUIController` interface needs both `enableLiveVideoUI` and `enableLiveVODUI`? Isn't `enableLiveVideoUI`  enough? 
It seems to me that `enableLiveVODUI` should be removed the `PlayerUIController` interface, and used only internally.
3. `videoDurationValue` should be a float.
4. Could you please send the PR on the dev branch instead of master?

rajeefmk:
Thanks @PierfrancescoSoffritti :)  I've added replies to your points below : 

1. I'd like the XML file defining the UI of the player to not be changed. If i'm not mistaken the only UI addition needed to use your improved live UI is the red/gray dot in the LIVE textview, which is added as a drawable on the textview. Therefore no change should be required on the XML file, am I correct?

_You're right. It won't affect my work. The only UI addition is the red/gray dot. Apart from that there's also the drawable padding that's added in the xml. Since the drawable itself is being added from the Java code, I can move the padding also to the same place. 
That said, I'm attaching screenshots after reverting back to default UI. Only concern is related to the seekbar (now that it's visible always) as you can see below, it looks like there is still spacing left on the right and left. Do let me know what you think about it._

live_stream_default
live_vod_stream_default

2. Why does the PlayerUIController interface needs both enableLiveVideoUI and enableLiveVODUI? Isn't enableLiveVideoUI enough?  It seems to me that enableLiveVODUI should be removed from the PlayerUIController interface, and used only internally.

_True. We could potentially handle it internally. Although, since we allow the change of seekbar progress via code by calling seekto() (instead of user manually moving it), I came across the use case where when this happens,  this method will help enable VOD mode for such cases if need be.  The alternative is to manage this in the onProgressChanged method. I refrained from that because -
a)  It gets called for every position and the UI will get keep redrawn
b) We wouldn't know the logic behind why the manual seekTo() was called in the parent code.
What are your thoughts on it ?_

4.  videoDurationValue should be a float.
_Sure. Will make the change._

5. Could you please send the PR on the dev branch instead of master?
_Sure. Will do. But let's please  finalize the discussion about above points here and then I'll make changes to new PR and give it to dev branch._

PierfrancescoSoffritti:
1. The seekbar is inside a LinearLayout, with weight 1, therefore it should automatically fill all the available space. If it doesn't, don't worry about it. Just revert to the original XML file and I'll take care of the rest.

2. I see your point. What about using `handleProgressChangedDuringLive` like this:

```
private void handleProgressChangedDuringLive(SeekBar seekBar, int i) {
  // ...
  if (i == videoDurationValue) {
    enableLiveVideoUI(true);
    return;
  } else {
    enableLiveVODUI(ture);
  }
  // ...
}
```
This would remove the need to manually call `enableLiveVODUI` as it will always be called automatically when needed.

It's quite important not to add extra noise to the interfaces. `PlayerUIController` is already quite messy on its own.

PierfrancescoSoffritti:
Hi, do you know when will you be able to send in the PR? I have to make some big changes for version 8, would be easier to accept your PR before making them.

rajeefmk:
@PierfrancescoSoffritti Sorry I couldn't reply earlier. Thanks for your suggestion regarding the enabling code for Live VOD mode. I'll go ahead with that. I should be able to update the PR by end of day tomorrow (25th June).

PierfrancescoSoffritti:
No worries. In the meantime i've made significant changes to the dev branch, namespaces and modules have different names now.

I suggest making a new fork of the dev branch and applying the changes for your PR there. Let me know if you have any issue :)

rajeefmk:
@PierfrancescoSoffritti I tried making a new fork from dev branch doing a build on the core-sample-app. I was hit with multi dex issue and the build failed. Are you planning to add support for multidex and enabling it ? I can take that up if need be.

PierfrancescoSoffritti:
No, no need for multidex. Clean the project and then run the sample app. It should work, let me know if it doesn't.

rajeefmk:
@PierfrancescoSoffritti Got the error again. Attaching stacktrace : 

```
Cannot fit requested classes in a single dex file. Try supplying a main-dex list.
# methods: 73637 > 65536
Message{kind=ERROR, text=Cannot fit requested classes in a single dex file. Try supplying a main-dex list.
```

Apparently it has crossed the max dex limit methods. Let's do one thing. Let this feature not be a blocker for the 8.0 release. Please go ahead with it. I'll probably fork from master once you've released 8.0

PierfrancescoSoffritti:
uhm, weird. I am able to build and run both sample apps, no dex issues for me.

Ok, the new version should be out by the end of this week!

PierfrancescoSoffritti:
which version of android are you using?

PierfrancescoSoffritti:
Problem solved. Feel free to work on the dev branch when you have time :)

rajeefmk:
@PierfrancescoSoffritti Superb. I've created a new pull request here https://github.com/PierfrancescoSoffritti/Android-YouTube-Player/pull/155 . 

Closing this one.

Issue #132, opened 2018-05-10
Add double tap to go forward/backward in DefaultPlayerUIController


PierfrancescoSoffritti:

rajeefmk:
@PierfrancescoSoffritti 
If you notice the youtube app, its buggy most of the time since half the time, I wasn't sure where to tap and until the tap is made, the player also doesn't come up. 

1. Are we going with similar approach of double tap on screen ? 
2. Or should we introduce the -10 & +10s icons to the left and right of the play/pause button ? (Similar to what Amazon prime does ?

PierfrancescoSoffritti:
Solution number 2 is basically already implemented, using `setCustomAction1` and `setCustomAction2`. It's just a matter of doing seekTo(time +- 10). Users of the library can take care of that.

I think solution 1 is more interesting. It shouldn't be too hard to achieve, all is needed are 2 transparent views that cover one half of the screen each. These views can then be used to detect double taps. 

I think the trickiest thing to implement (but i may be wrong) may be the detection of multiple double taps. eg: if the user taps 4 times on the right view, the player should seek to currentTime + 20 instead of currentTime + 10.

rajeefmk:
Ha. That's interesting. I just checked out the latest youtube app and the double tap issue seems to be fixed and it's working beautifully. 

Also, about the multiple double tap, I think once the first double tap activates the fast forward (backward) mode, then the player doesn't differentiate between single/double tap. I was able to increment in single 10s counts also. 

Let's try and emulate the exact behavior of the native app.  Would be happy to take this up.

PierfrancescoSoffritti:
You're right. After the first two taps it's only one to skip 10 seconds.

Sounds great to me, thanks for helping :)

abahl817:
Hey @PierfrancescoSoffritti, I have an issue related to the custom actions. I set the actions in initlistener method using the YouTubePlayerTracker class. The buttons appear at first when the video player is initializing but when the video starts to play the buttons disappear. This is the code I used, where ytVideo is the YouTubePlayerView: `
ytVideo.initialize(new YouTubePlayerInitListener() {

                    @Override
                    public void onInitSuccess(@NonNull final YouTubePlayer youTubePlayer) {
                        final YouTubePlayerTracker tracker = new YouTubePlayerTracker();
                        youTubePlayer.addListener(tracker);
                        PlayerUIController playerUIController = ytVideo.getPlayerUIController();
[...]

Issue #71, opened 2017-12-19
Add support for 18:9 aspect ratio


PierfrancescoSoffritti:
Add support for 18:9 aspect ratio, in the same way the official YouTube app does.

PR #155, opened 2018-06-29
feat: Introduce support for Live-VOD mode for live streams

rajeefmk:
The changes included are :

- Introduced Live-VOD (Video on Demand) mode similar which is enabled when user moves the seekbar during live stream.
- Live stream indicator tag is made clickable and now has the same functionality as that of native app wherein it'll restore the live state.
- Seekbar is always visible now and during Live stream, it's progress position value will be that of the video duration.

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/ui/DefaultPlayerUIController.java
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/ui/DefaultPlayerUIController.java
@@ -71,6 +71,10 @@ public class DefaultPlayerUIController implements PlayerUIController, YouTubePla
     private boolean showPlayPauseButton = true;
     private boolean showBufferingProgress = true;
 
+    private boolean isLive = false;
+    private boolean isLiveVOD = false;
+    private float videoDurationValue = 0;
+
     public DefaultPlayerUIController(@NonNull YouTubePlayerView youTubePlayerView, @NonNull YouTubePlayer youTubePlayer) {
         this.youTubePlayerView = youTubePlayerView;
         this.youTubePlayer = youTubePlayer;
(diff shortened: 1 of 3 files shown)


PierfrancescoSoffritti:
Great, thanks! expect it available in v8 soon :)

PierfrancescoSoffritti:
If you'd like to take on other issues just let me know!

rajeefmk:
@PierfrancescoSoffritti Absolutely.

PierfrancescoSoffritti:
[issue 132]( https://github.com/PierfrancescoSoffritti/Android-YouTube-Player/issues/132)
this one is similar to the one you just sent

[issue 71](https://github.com/PierfrancescoSoffritti/Android-YouTube-Player/issues/71)
this one is also about implementing features from the YouTube app, but probably a bit more challenging

rajeefmk:
@PierfrancescoSoffritti What a coincidence ! I was actually going through this https://github.com/PierfrancescoSoffritti/Android-YouTube-Player/issues/132 very same ticket. 

Continuing the discussion there.

PierfrancescoSoffritti:
I've been testing the new live controls, it's really hard to ensure consistency. The available buffer for some live videos appears to be of 2 hours, for other is different. Therefore is basically impossible to map correctly the video's duration on the seekbar.
Almost all the time, when manually seeking to the start or end of the seekbar the behavior is inconsistent.

I have been trying to find a solution to these problems but I haven't been able to. Do you have any ideas? If it's not possible to solve these issues I'm afraid I will have to revert the changes :/

rajeefmk:
@PierfrancescoSoffritti You're absolutely right. So the seekbar behaviour is directly dependent on how on the total duration of the live session and not precisely when the actual broadcast started. This will definitely create inconsistency. 

I think you'll have to go ahead revert the changes :/ Sorry about that !

PierfrancescoSoffritti:
No worries, it is quite tricky actually. The buffer is behaving like a sliding window, not sure yet how we could implement it.

Issue #231, opened 2018-10-21
Manifest merger failed 


naveenthontepu:
In my application i want use `android:allowBackup="false"` but i getting a build error saying 

```
Manifest merger failed : Attribute application@allowBackup value=(false) from AndroidManifest.xml:9:9-36
	is also present at [com.pierfrancescosoffritti.androidyoutubeplayer:core:8.0.1] AndroidManifest.xml:15:9-35 value=(true).
	Suggestion: add 'tools:replace="android:allowBackup"' to <application> element at AndroidManifest.xml:7:5-52:19 to override.
```

I can simply add `tools:replace="android:allowBackup"`. But why is your library using `allowBackup`?

PierfrancescoSoffritti:
Hey, thanks for reporting it. It was just automatically added by AndroidStudio and i forgot to remove it.

Will be removed in the next version.

PR #241, opened 2018-10-29
Removed Unneeded tag in the core AndroidManifest

jguerinet:
The application tag (and its accompanying app name) were not used, and when used within an app that had different values for `supportsRtl` and `allowBackup` would cause the Manifest merger to fail. 
I also fixed a small typo within the README (Android Studio removed a few empty whitespaces when committed).


jguerinet:
Fixes #231

PierfrancescoSoffritti:
Looks good!! thank you for the help :)

PR #244, opened 2018-10-31
Fixed background playback by adding it as an option

CajetanP:
After doing some research I found out that video background playback stopped working because somewhere along the way Google implemented the Visibility API for WebView.

This fix basically lets users enable the functionality with a simple method call before initialising the PlayerView.

If it's enabled, if the visibility changes to invisibile (user switches to another app or home screen) while the video is playing, the library overrides the default behaviour so that the WebView is treated as if it was still visible, thus the video keeps playing in the background. Of course playback is still tied to the activity so if it gets destroyed it'll stop but that's up to the user to handle I suppose.


PierfrancescoSoffritti:
Hey, thanks a lot for investigating! I have a few requests: 

1. Could you send the pull request to the dev branch instead of the master branch?

2. If the user has to enable background playback before calling the initialize function, we should just pass an extra `boolean enableBackgroundPlayback` to the initialize function, instead of adding the method `enableBackgroundPlayback`.

3. The `isPlaying` variable in `YouTubePlayerView` doesn't seem necessary and it has the side effect of duplicating the state of the player. Right now the only source of true for the state of the player is the player itself, this variable breaks that.
The `onVisibilityChanged` method should be: 
```java
@Override
protected void onWindowVisibilityChanged(int visibility) {
  if (!backgroundPlaybackEnabled)
    super.onWindowVisibilityChanged(visibility);
}
```



What do you think?

CajetanP:
1. Absolutely, no problem. 
2. That works too, I'll do that. 
3. Well yes, I don't really like that variable. I added it because I though the visibility should be overridden only if the video is playing, otherwise it might be better to let the default behaviour stay for optimalisation or something, I don't really know what Android does with an invisible view behind the scenes. The variable is here because I don't quite now how to get the state of the player from inside the player, if you have an idea I'll gladly get rid of it. 
4. Sure, sorry about those, I was in a bit of a hurry. Totally my bad.

PierfrancescoSoffritti:
3. I think for now we can just remove the variable, as I suggested in the other comment. Since we are already overriding the method for when the player is playing. If there's an issue with overriding it we would have problems anyway.

CajetanP:
Okay, I made the adjustments you asked for. Had to fix all the examples as well because now that the option is in initiaize they wouldn't build otherwise but it wasn't much of a problem.

PierfrancescoSoffritti:
Hi, just tested. It doesn't work. You need to apply these changes:

```java
@Override
protected void onWindowVisibilityChanged(int visibility) {
  if (backgroundPlaybackEnabled && (visibility == View.GONE || visibility == INVISIBLE))
    return;
        
  super.onWindowVisibilityChanged(visibility);
}
```

CajetanP:
Oh right, sorry for that. I seem to have refactored my code that was working and because of that refactoring it all of a sudden stopped working at all. Quite weird but nevermind, I amended the commit with the working version and this time tested it once again before pushing.

PierfrancescoSoffritti:
Good!
This code is quite confusing and can lead to problems
```java
@Override
protected void onWindowVisibilityChanged(int visibility) {
        
  if (!backgroundPlaybackEnabled) {
    super.onWindowVisibilityChanged(visibility);
    return;
  }

  // Override for background playback
  if (visibility != View.GONE)
    super.onWindowVisibilityChanged(View.VISIBLE);
  }
}
```

could you change it to the one I suggested? 

```java
@Override
protected void onWindowVisibilityChanged(int visibility) {
  if (backgroundPlaybackEnabled && (visibility == View.GONE || visibility == INVISIBLE))
    return;
        
  super.onWindowVisibilityChanged(visibility);
}
```

CajetanP:
Are you sure it's a good idea to treat a View that's **GONE** as if it were **VISIBLE**? As far as I'm aware, the situation we're interested in, which is when the player is there somewhere but the user can't see it directly is represented by the **INVISIBLE** state. Gone means it doesn't even take up space in the layout. Now I'm not sure under which circumstances the state turns to gone, or if it does at all for that matter but just in case I'd leave only `visibility == View.INVISIBLE` in the if statement.
I'll test the code you suggested tomorrow.

PierfrancescoSoffritti:
Ok! if I remember correctly it doesn't work if the check is only on `VIEW.INVISIBLE`. But let me know, I may be wrong!

CajetanP:
Well you were right, apparently with your version of the code the check on View.GONE is also necessary, strange but I guess the important thing is that it works. Since possible states are just VISIBLE, INVISIBLE and GONE I replaced that 'or' with `visibility != VISIBLE` becuase it's equivalent and just better visually.

PierfrancescoSoffritti:
Great! thanks a lot!

PR #251, opened 2018-11-09
Allow customizing iframe parameters using new IFramePlayerOptions class

matthewrkula:
### Problem
I need to be able to specify certain YouTube iframe parameters which are listed out [here](https://developers.google.com/youtube/player_parameters). Right now those are hardcoded in the `youtube_player.html` file. 

### Solution
I added an optional new class called `IFramePlayerOptions` where you can specify these parameters. I have kept the defaults as what the hardcoded values were before. This class is injected into the html string where the `<<playerVars>>` is. If there is a cleaner way to inject the values I am happy to update the PR.


PierfrancescoSoffritti:
Hey, this looks good! thanks a lot for the help! It makes much more sense to have access to the player params.

I have only one request, could you make the pull request on the dev branch? then it's good to go :)

PierfrancescoSoffritti:
Great! thank you :) I will release the new version of the library this week 👍

PR #254, opened 2018-11-13
Hide title, watch later, share and YouTube watermark

marcorighini:
This is a fix for hiding the unwanted top bar buttons (watch later, share) and title. This also removes lower right YouTube watermark. In this way you have a clean webview.

### DIFF
--- a/core/src/main/res/raw/youtube_player.html
+++ b/core/src/main/res/raw/youtube_player.html
@@ -45,7 +45,11 @@
     			width: '100%',
     			
                 events: {
-    				onReady: function(event) { YouTubePlayerBridge.sendReady() },
+    				onReady: function(event) {
+    				    document.querySelector("#youTubePlayerDOM").contentDocument.querySelector('.ytp-chrome-top').style.display = "none"
+    				    document.querySelector("#youTubePlayerDOM").contentDocument.querySelector('.ytp-icon-watermark').style.display = "none"
+    				    YouTubePlayerBridge.sendReady()
+    				},
     				onStateChange: function(event) { sendPlayerStateChange(event.data) },
     				onPlaybackQualityChange: function(event) { YouTubePlayerBridge.sendPlaybackQualityChange(event.data) },
     				onPlaybackRateChange: function(event) { YouTubePlayerBridge.sendPlaybackRateChange(event.data) },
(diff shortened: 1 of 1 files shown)


PierfrancescoSoffritti:
Hi, interesting. Can this be used to remove the title as well? 

My only concern is that it's a bit hacky. But I like it, so I'm ok merging it. I will probably make some changes to make it optional though.

Could you please send the pr on the dev branch instead of master?

marcorighini:
The solution is a bit hacky but I think is the only solution we have at the moment.
I've updated the pr description ;) Let me know if you think is better to have more granular functions for removing components one by one

PierfrancescoSoffritti:
> Let me know if you think is better to have more granular functions for removing components one by one

I can do that, no problem!

I also wonder if this is against YouTube terms of use hehe will definitely have to add it as optional.

I will test it tomorrow, if everything is fine I'll accept the pr 👍

matthewrkula:
Is this going to violate some sort of ToS with YouTube embeds? Based on the August 23rd revision history [here](https://developers.google.com/youtube/player_parameters#showinfo), they purposefully deprecated `showinfo` to 
> make the YouTube viewing experience consistent across different platforms 

If this is going to be a valid option to replace the YouTube Android SDK I don't know if Google is going to let this fly.

zafarsk227:
Can we Remove the "More videos" also? It is creating a bad image of my app to the users as they cant click on those buttons(as I don't want them to do so).They are thinking this is a Bug.

PierfrancescoSoffritti:
Hey, sorry for the late reply!
I decided not to merge the pull request, is better to not introduce hacky solutions in the library. If someone needs to use them they can fork the library and implement the changes on their own fork.

In the next few days i'm going to release version 10.0.0, it will be possible to enable the web ui instead of the native ui. Using the web ui those buttons will be clickable.

mayank2212:
Hello @marcorighini watermark is not getting removed by using this, can you help me with that

palatirco:
Hello, how can i modify the project to implement the changes of @marcorighini ?

I tried download, modify code, and import the youtube player in my project but i cant make it work.

i received 3 errors:

ERROR: Unable to resolve dependency for ':app@debug/compileClasspath': Could not resolve project :android-youtube-player.

ERROR: Unable to resolve dependency for ':app@debugAndroidTest/compileClasspath': Could not resolve project :android-youtube-player.

ERROR: Unable to resolve dependency for ':app@debugUnitTest/compileClasspath': Could not resolve project :android-youtube-player.

Thanks

PR #280, opened 2019-01-09
Customise Play/Pause button's Drawable and OnClickListener

vibin:


vibin:
After this change, the red YouTube play icon seems to overlap with this library's player icon...trying to figure out why.

PierfrancescoSoffritti:
Hi, thanks for the pull request!

I don't think that adding a new View for the play/pause button is a good idea. Users that want to change the ui can do it by building a custom ui.

Just setting an onclick listener should be enough. What do you think?

vibin:
It's not a new View, play/pause button view already exists, right?

Regarding customisation, I actually tried using inflateCustomPlayerUI(), but there's a lot of useful functionality in DefaultPlayerUIController which I had to copy-paste: including fade-in, fade-out, scrim, progress, so on.

I can't set OnClickListener for the play/pause button specifically, so a non-trivial way to do it would be to set OnTouchListener on the YouTubePlayerView itself and do MotionEvent math to detect that play/pause button is clicked.

PierfrancescoSoffritti:
> It's not a new View, play/pause button view already exists, right?

It's a new drawable. If we add that functionality for the play/pause button then why not adding it for every other button? it goes out of hand pretty quickly.

> Regarding customisation, I actually tried using inflateCustomPlayerUI(), but there's a lot of useful functionality in DefaultPlayerUIController which I had to copy-paste: including fade-in, fade-out, scrim, progress, so on.

Yeah, the code of `DefaultPlayerUIController` is not very well written. Instead of adding extra stuff the right thing to do is to refactor the class and see if the refactored code can be reused when building custom uis.

> I can't set OnClickListener for the play/pause button specifically, so a non-trivial way to do it would be to set OnTouchListener on the YouTubePlayerView itself and do MotionEvent math to detect that play/pause button is clicked.

One simple way to do it is just to add a new click listener to the UI controller and then manually call it when the button is clicked. But again, not sure if it makes sense to add this option. It's the same problem I have with the drawable, why not adding it for every other button?

The best thing to solve everything in a nice way is to refactor `DefaultPlayerUIController`.

vibin:
Sadly, I agree the right thing to do is to refactor `DefaultPlayerUIController`.

But till then, is there any way I can reuse code from `DefaultPlayerUIController` in a custom controller? I can't use inheritance because the class doesn't seem to be designed for it, nor can I use delegation.

PierfrancescoSoffritti:
For now you could just make a local version of it (copy-paste the code in your own class) and apply there the changes you need. Then you can use it as a controller for your custom UI.

PierfrancescoSoffritti:
Yeah, I hope I won't have to wait to long before doing some more work on the lib :)

In the mean time I've opened the issue for it.

PierfrancescoSoffritti:
Issue #281 has been closed, a few components have been refactored out of `DefaultPlayerUIController`. They will be available starting from version 10.0.0 in the upcoming days :)

vibin:
thank you so much!

PR #329, opened 2019-03-23
Hides youtube watermark and topbar buttons

sarbagyastha:
This technique virtually maximizes the viewport so that everything except the video in shrinked to small size(i.e. watermark and buttons become so small that it seems to be not visible.) [Live Demo](https://sarbagyastha.com.np/youtubeplayer.html) 

Merits: 
- Since we aren't hiding anything( just making it small 😉) so it won't be violation of YouTube's TOS.
- Video quality is increased as the iframe player automatically adjust video quality as per the viewport size.


PierfrancescoSoffritti:
Hey, thanks for the PR!
Unfortunately this looks very much like a hack, I think it wouldn't be good to implement this in the library for two reasons:
1. Being a hack YouTube may very well identify this as illicit use of the API.
2. Always downloading videos at the highest possible resolution isn't a desired behavior.

Let me know what are your thoughts about this :)

Issue #326, opened 2019-03-20
Playing the video in background while activity is stopped


abelhamilton:
Hello, thanks again for this wonderful library!

Maybe I'm missing something but the player continues playing in the background when I press the home button or open another activity. It happens sometimes only when the video is not loaded yet even when mYouTubePlayer.pause(); is called in the onPause method.

Here is what I do :
```
@Override
    protected void onCreate(Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        setContentView(R.layout.activity_video);
        ButterKnife.bind(this);

        mVideo = (Video) getIntent().getSerializableExtra(Constant.INTENT_OBJECT);
        mVideoActivity = (VideoAct) getIntent().getSerializableExtra(Constant.INTENT_ACTIVITY);

        initYoutubePlayer();
    }

private void initYoutubePlayer() {
        getLifecycle().addObserver(youtubePlayerView);
        youtubePlayerView.addYouTubePlayerListener(new AbstractYouTubePlayerListener() {
            @Override
            public void onReady(@NonNull YouTubePlayer youTubePlayer) {
                mYouTubePlayer = youTubePlayer;
                int startTime = 0;
                if (mVideoActivity != null) {
                    startTime = mVideoActivity.timeCode;
                }
                YouTubePlayerUtils.loadOrCueVideo(mYouTubePlayer, getLifecycle(), videoKey, startTime);
            }
        });
}

@Override
public void onPause() {
    super.onPause();
    if (mYouTubePlayer != null) {
       mYouTubePlayer.pause();
    }
}

@Override
protected void onDestroy() {
    super.onDestroy();
    youtubePlayerView.release();
}
```
Thanks in advance

PierfrancescoSoffritti:
mmm this snippet looks good, you actually don't need to override onPause and onDesotry, but that's not the issue.

Are these the only places where you're using `mYouTubePlayer`?

I need more details to understand what's wrong.

abelhamilton:
Yes, that's the only places where I'm using `mYouTubePlayer`.
I'm using : implementation 'com.pierfrancescosoffritti.androidyoutubeplayer:core:10.0.1'

It happens quite randomly when I quit when the youtube logo is about to appear or have just appeared and the video is still loading.

Which other details can I give you?

PierfrancescoSoffritti:
Are you using the player in a Fragment or in an Activity?
[...]

PR #345, opened 2019-04-08
Fix issue with stopping playback

hannojg:
When having enableBackgroundPlayback enabled the player stops sometimes when the activity gets into the background. Basically, if background playback is enabled we don't need to mind whether the activity is in fore- or background. This is a fix for it.

I currently can't compile the library on my own, so I haven't tested the changes. I assume it should be fine this way. This issue was raised since fix of issue #326

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/LegacyYouTubePlayerView.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/LegacyYouTubePlayerView.kt
@@ -4,7 +4,6 @@ import android.content.Context
 import android.content.IntentFilter
 import android.net.ConnectivityManager
 import android.util.AttributeSet
-import android.util.Log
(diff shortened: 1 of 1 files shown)


PierfrancescoSoffritti:
Hey, thanks! could you please send the PR to the dev branch instead of master? I'll try to test it during the weekend :)

PierfrancescoSoffritti:
Hey, sorry for taking such a long time to review the PR. Just changed town and job, busy times :) thanks a lot for the help! I will release the next version some day next week!

devbest555:
> When having enableBackgroundPlayback enabled the player stops sometimes when the activity gets into the background. Basically, if background playback is enabled we don't need to mind whether the activity is in fore- or background. This is a fix for it.
> 
> I currently can't compile the library on my own, so I haven't tested the changes. I assume it should be fine this way. This issue was raised since fix of issue #326

I tried your code in my side but same error yet.
Youtube video stop playing in background sometimes.
Looking forward to chat with you ASAP.

PR #390, opened 2019-07-09
added setup in IFramePlayerOptions for setting up video's origin

kpiskur:


core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
```diff
@@ -7,7 +7,7 @@ import org.json.JSONObject
  * Options used to configure the IFrame Player. All the options are listed here:
  * [IFrame player parameters](https://developers.google.com/youtube/player_parameters#Parameters)
  */
-class IFramePlayerOptions private constructor(private val playerOptions: JSONObject) {
+class IFramePlayerOptions private constructor(private val playerOptions: JSONObject, val origin: String) {
```
PierfrancescoSoffritti: Why is origin passed to the constructor? We should keep using the builder pattern instead of doing this.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
```diff
@@ -87,6 +87,16 @@ class IFramePlayerOptions private constructor(private val playerOptions: JSONObj
             return this
         }
 
+        /**
+         *  If you are using the IFrame API, which means you are setting the enablejsapi parameter value to 1, you should always specify your domain as the origin parameter value.
+         * @param origin - provides an extra security measure for the IFrame API and is only supported for IFrame embeds.
+         */
+        fun origin(origin: String): Builder {
+            addInt(ENABLE_JS_API, 1)
```
PierfrancescoSoffritti: No need to set ENABLE_JS_API here.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
```diff
@@ -87,6 +87,16 @@ class IFramePlayerOptions private constructor(private val playerOptions: JSONObj
             return this
         }
 
+        /**
+         *  If you are using the IFrame API, which means you are setting the enablejsapi parameter value to 1, you should always specify your domain as the origin parameter value.
```
PierfrancescoSoffritti: Align the documentation of this methods with the others.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
kpiskur: Do you think adding method in _IFramePlayerOptions_ instead of that would work?

`  fun getOrigin(): String {
        return playerOptions.getString(Builder.ORIGIN)
    }`

kpiskur: Do you think adding method in _IFramePlayerOptions_ instead of that would work?

`  fun getOrigin(): String {
        return playerOptions.getString(Builder.ORIGIN)
    }`

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
```diff
@@ -17,13 +17,17 @@ class IFramePlayerOptions private constructor(private val playerOptions: JSONObj
...
+    }
+
     class Builder {
         companion object {
             private const val AUTO_PLAY = "autoplay"
             private const val CONTROLS = "controls"
             private const val ENABLE_JS_API = "enablejsapi"
             private const val FS = "fs"
-            private const val ORIGIN = "origin"
+            const val ORIGIN = "origin"
```
PierfrancescoSoffritti: Why is this not private?

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/LegacyYouTubePlayerView.kt
```diff
@@ -15,21 +15,21 @@ import com.pierfrancescosoffritti.androidyoutubeplayer.R
...
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.options.IFramePlayerOptions
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.FullScreenHelper
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.NetworkListener
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.PlaybackResumer
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.ui.DefaultPlayerUiController
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.ui.PlayerUiController
-import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.NetworkListener
 
-internal class LegacyYouTubePlayerView(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0):
+internal class LegacyYouTubePlayerView(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0) :
```
PierfrancescoSoffritti: Why are there still changes to this file? It has nothing to do with this pull request.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
kpiskur: because in other case I couldn't access ORIGIN constant from _IFramePlayerOptions_ to get origin value from _playerOptions_

PR #471, opened 2019-11-07
Fix webview in 5. ( lollipop) crash 

cuong292:
-Fix WebViewYouTubePlayer crash on Android 5 
(https://stackoverflow.com/questions/41025200/android-view-inflateexception-error-inflating-class-android-webkit-webview/41721789#41721789).
- I also tryed to downgrade appcompat version but it still not work on my 5. device.and this solution work well for all.


PierfrancescoSoffritti:
Hey, thanks for the pull request but this solution looks a bit hacky. And it's not good practice to implement hacky solutions in a library used by many people.

Issue #419, opened 2019-08-21
Error inflating class com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView


Hasan-Mahmud:
I got Error:
android.view.InflateException: Binary XML file line #24: Error inflating class com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView
![p](https://user-images.githubusercontent.com/40826552/63406423-a6955680-c40b-11e9-8e0f-2c013c93a516.JPG)

I use 'com.pierfrancescosoffritti.androidyoutubeplayer:core:10.0.3'

Yagnesh123456:
i got same error when xml file without videoId

PierfrancescoSoffritti:
[It doesn't make sense to set autoplay to true without providing a video id](https://github.com/PierfrancescoSoffritti/android-youtube-player#autoplay).

Hasan-Mahmud:
> 
> 
> i got same error when xml file without videoId .if you want solve this error then set autoplay false or set video id in xml file

Thanks for your solution. It works now.

SergioFuchs:
I´m having the same problem with API 21, on API 27, 28 the problem doesn´t happen. Only on old ones ... 

 <com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView
                android:id="@+id/youtube_player_view"
                android:layout_width="match_parent"
                android:layout_height="wrap_content"
                app:videoId="I08i3ztnJdE"
                app:autoPlay="false"
                app:showFullScreenButton="true" />

PierfrancescoSoffritti:
I doubt this is related to API level. Are you sure this is the issue you're having?

SergioFuchs:
> I doubt this is related to API level. Are you sure this is the issue you're having?

I´ts a very simple page and it´s working on API 27, 28 simulator and real devices.
But if open this Activity on API 21 I got the error above. 
If I remove the YouTubePlayerView from the Activity it opens with no erros.
I´m already using Androidx library
Thanks

PierfrancescoSoffritti:
Hey, I'm currently traveling and don't have my personal laptop with me. I can try fixing #430 today but won't have time for #491 (at least for the next week). Feel free to send pull requests.

SJOwl:
```
internal class WebViewYouTubePlayer constructor(
    context: Context,
    attrs: AttributeSet? = null,
[...]

PR #473, opened 2019-11-09
use application context for WebView initialization

SJOwl:
Issue https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/419 fix


PierfrancescoSoffritti:
How do you know that this solves the issue? is there a way for me to replicate it?

Issue #461, opened 2019-10-25
App crashes on Android 5 - Resources$NotFoundException: String resource ID #0x2040002


saurabhdtu:
The crash is continuously happening for android lollipop micromax and intex devices
here is the stacktrace

com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.WebViewYouTubePlayer.<init> + 25 (WebViewYouTubePlayer.java:25)
--
  | com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.WebViewYouTubePlayer.<init> + 24 (WebViewYouTubePlayer.java:24)
  | com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.LegacyYouTubePlayerView.<init> + 34 (LegacyYouTubePlayerView.java:34)
  | com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.LegacyYouTubePlayerView.<init> + 31 (LegacyYouTubePlayerView.java:31)
  | com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView.<init> + 30 (YouTubePlayerView.java:30)
  | com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView.<init> + 27 (YouTubePlayerView.java:27)

Please fix it asap

PierfrancescoSoffritti:
This is not very helpful. Let me know how to replicate this issue. Do I just need to run the sample app on any device running Android 5?

saurabhdtu:
Yeah. 
Just run this code 
`YouTubePlayerView youtubePlayerView = new YouTubePlayerView(getContext());`
intex, micromax devices with api level less than 22.
Intex - AquaAce mini
Micromax - Canvas 5, Canvas Spark, Canvas juice2

anthony3444:
Hi!

I have the same problem on Android 5.1.1 (HUAWEI Y6) : 
```
Caused by android.content.res.Resources$NotFoundException
String resource ID #0x3040005
```

Stacktrace:

```
android.content.res.HwResources.getText (HwResources.java:1299)
android.webkit.WebView.<init> (WebView.java:472)
com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.WebViewYouTubePlayer.<init> (WebViewYouTubePlayer.kt:25)
[...]

PR #476, opened 2019-11-11
use application context for WebView initialization

SJOwl:
To know it works: open app on android-21 emulator. Without it we have Inflation Error.


PierfrancescoSoffritti:
Ok, I'll try it out. 

Just for reference, this pull request is about issue #419.

PierfrancescoSoffritti:
So this seems to be an issue with androidx.appcompat version 1.1.0. I think the best solution is to wait for the Android team to fix it.

In the meantime [this is a temporary solution](https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/461#issuecomment-550050683).

PR #517, opened 2020-02-09
Fix #105 click listeners test

### LINKED ISSUE #105: click listeners do not work
The click events are never fired. If I put a transparent view in front of my youtube view, the click event is fired for that transparent view, but then that transparent view steals the click event from the youtube seekbar, play, fullscreen button etc. 

Any suggestions? 

```
youTubePlayerView.setOnTouchListener(new View.OnTouchListener() {
            @Override
            public boolean onTouch(View v, MotionEvent event) {
                log.debug( "youTubePlayerView:MOTIONEVENT");
                // Get motion event type
                switch (event.getAction()) {
                    // This event is when user clicks down
                    case MotionEvent.ACTION_DOWN:
                        log.debug( "youTubePlayerView:ACTION_DOWN");
                        EventBus.getDefault().post(new VideoUIControlEvent("EXPAND_VIDEO"));
                        break;
                }
                return true;
            }

        });
```

 ```
       youTubePlayerView.setOnClickListener(new View.OnClickListener() {
            @Override
            public void onClick(View v) {
                log.debug( "youTubePlayerView:setOnClickListener");
            }
        });
```

### PR DESCRIPTION
**Pull request by AbelTesfaye**

### CODE REVIEW

File: core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt [hunk 1]
```diff
@@ -16,11 +17,51 @@ open class SixteenByNineFrameLayout: FrameLayout {
     constructor(context: Context, attrs: AttributeSet?) : this(context, attrs, 0)
     constructor(context: Context, attrs: AttributeSet? = null, defStyleAttr: Int = 0) : super(context, attrs, defStyleAttr)
 
+    var mCustomTouchListener: OnTouchListener? =null
```
PierfrancescoSoffritti: Don't use hungarian notation, customTouchListener is fine.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt
```diff
@@ -16,11 +17,51 @@ open class SixteenByNineFrameLayout: FrameLayout {
...
     override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
         if (layoutParams.height == ViewGroup.LayoutParams.WRAP_CONTENT) {
             val sixteenNineHeight = View.MeasureSpec.makeMeasureSpec(View.MeasureSpec.getSize(widthMeasureSpec) * 9 / 16, View.MeasureSpec.EXACTLY)
             super.onMeasure(widthMeasureSpec, sixteenNineHeight)
         } else
             super.onMeasure(widthMeasureSpec, heightMeasureSpec)
     }
+
+    override fun onInterceptTouchEvent(ev: MotionEvent): Boolean {
+        /*
```
PierfrancescoSoffritti: The goal of this comment is to document the method, not implementation details. Instead of a comment inside the method, put kdoc on the method. Further reading [here](https://kotlinlang.org/docs/reference/kotlin-doc.html#kdoc-syntax).

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt
```diff
@@ -16,11 +17,51 @@ open class SixteenByNineFrameLayout: FrameLayout {
...
+            children handle the event or not
+
+            *reference: https://developer.android.com/training/gestures/viewgroup#intercept
+
+         */
+
+
+        // Even though click and long click events are being called, the touch listener
+        // doesn't seem to work. So we need to register and use our own touch listener
+       val isTouchEventHandled: Boolean? = mCustomTouchListener?.onTouch(rootView,ev)
```
PierfrancescoSoffritti: val isTouchEventHandled = customTouchListener?.onTouch(rootView, ev) ?: false

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt
```diff
@@ -16,11 +17,51 @@ open class SixteenByNineFrameLayout: FrameLayout {
...
+
+        if(isTouchEventHandled == null || !isTouchEventHandled)
+            onTouchEvent(ev)
+
+        return super.onInterceptTouchEvent(ev)
+    }
+
+    override fun setOnTouchListener(l: OnTouchListener?) {
```
PierfrancescoSoffritti: Can nullability be remove? Also use a more descriptive name for this variable.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt
```diff
@@ -16,11 +17,51 @@ open class SixteenByNineFrameLayout: FrameLayout {
...
+        mCustomTouchListener = l;
+        super.setOnTouchListener(l) // this doesn't seem to register the touch listener
```
PierfrancescoSoffritti: What does this comment means? Why are we calling it if it doesn't work?

build.gradle
```diff
@@ -1,6 +1,7 @@
 apply from: './dependencies.gradle'
 
 buildscript {
+    ext.kotlin_version = '1.3.61'
```
PierfrancescoSoffritti: Instead of doing it here you should change the version in the file dependencies.gradle.

build.gradle
```diff
@@ -11,6 +12,7 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$versions.kotlin"
 
         classpath "com.novoda:bintray-release:$versions.bintrayPlugin"
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
PierfrancescoSoffritti: Why is did you add this line? isn't it the same as line 12?

PierfrancescoSoffritti:
Hi, I did a first pass without going into details. A few high level comments:

Each pull request should be as small as possible. It should be about a single logical change/new feature. This makes it way easier to review and reduces the likelihood of introducing bugs. 
This pull request does two things: 
1. sets up testing environment
2. fixes a bug (with tests)
I would suggest to split it into 2 separate pull requests.

Tests should not be in the sample-app module (unless they're testing the sample app), but in the module where the feature they are testing exists.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt
AbelTesfaye: Would removing nullability be a good way to go? Shouldn't the user be able to remove the listener by doing `myView.setOnTouchListener(null)` like on every other view?

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/SixteenByNineFrameLayout.kt
AbelTesfaye: I assumed it might have an effect later on. But I saw how it was implemented. It does nothing, so i've removed it now.

build.gradle
AbelTesfaye: 😅

AbelTesfaye:
> Hi, I did a first pass without going into details. A few high level comments:
> 
> Each pull request should be as small as possible. It should be about a single logical change/new feature. This makes it way easier to review and reduces the likelihood of introducing bugs.
> This pull request does two things:
> 
> 1. sets up testing environment
> 2. fixes a bug (with tests)
> 
> I would suggest to split it into 2 separate pull requests.
> 
> Tests should not be in the sample-app module (unless they're testing the sample app), but in the module where the feature they are testing exists.

I've made the requested changes. I've also split the PR in two, so I'm closing this one and re-opening 2 others.

PR #525, opened 2020-02-23
Fix single click wrong click exception

AbelTesfaye:
This exception is present in recent versions of Android.
Its not present on Android 7.1 Nougat (API 25).

### DIFF
--- a/core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/YouTubePlayerViewClickTest.kt
+++ b/core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/YouTubePlayerViewClickTest.kt
@@ -7,6 +7,7 @@ import androidx.test.espresso.Espresso.onView
 import androidx.test.espresso.action.ViewActions.*
 import androidx.test.espresso.matcher.ViewMatchers.withId
 import androidx.test.filters.LargeTest
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement.runOnUiThread
 import androidx.test.rule.ActivityTestRule
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.testActivity.TestActivity
@@ -40,7 +41,9 @@ class YouTubePlayerViewClickTest {
     fun testSingleClickListener() {
         // Prepare
         val mockClickListener = mock(View.OnClickListener::class.java)
-        youTubePlayerView.setOnClickListener(mockClickListener)
+        runOnUiThread {
+            youTubePlayerView.setOnClickListener(mockClickListener)
+        }
 
         // Act
         onView(withId(R.id.youtube_player_view)).perform(click())


core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/YouTubePlayerViewClickTest.kt
```diff
@@ -40,7 +41,9 @@ class YouTubePlayerViewClickTest {
     fun testSingleClickListener() {
         // Prepare
         val mockClickListener = mock(View.OnClickListener::class.java)
-        youTubePlayerView.setOnClickListener(mockClickListener)
+        runOnUiThread {
+            youTubePlayerView.setOnClickListener(mockClickListener)
+        }
```
PierfrancescoSoffritti: Why do we have to do this here and not everywhere else?

AbelTesfaye: This is the only failing portion. The others work without it.

PierfrancescoSoffritti: Do you know what's the reason? It seems odd.

AbelTesfaye: Nope, I have absolutely no clue. I even knew the bug existed when I tested on the emulator

PierfrancescoSoffritti: I think we should understand what's happening. It might be a symptom of a bigger problem.

AbelTesfaye: Looked everywhere, but no one else has encountered this issue :(

Should I file a bug on the android issue tracker?

PierfrancescoSoffritti: mmm if this is a bug on android it should be fairly common. I would say it's on our side. I'll have a look.

PierfrancescoSoffritti: Looks like threading is a bit strange in instrumentation tests. I don't have time to properly understand it right now. So I'd say we can leave it as it is.

When I run the tests SwipeDown and SwipeRight also fail. Are you aware of it?

AbelTesfaye: No, those didn't fail on the emulator and my physical device.

AbelTesfaye: So the `setOnLongClickListener()` is the only one that doesn't fail? This is strange

PierfrancescoSoffritti: Yeah, i had multiple tests failing. Let's submit this one and see what happens once we run tests in travis.

PR #532, opened 2020-03-06
Add tests for FullScreenHelperTest

AbelTesfaye:
Adds tests for classes in core utils


core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/FullScreenHelperTest.kt
```diff
@@ -0,0 +1,119 @@
...
+ * This class tests if the FullScreenHelper's enterFullScreen(), exitFullScreen(), toggleFullScreen(),
+ * addFullScreenListener() and removeFullScreenListener() methods work as expected
+ */
+@LargeTest
+class FullScreenHelperTest {
+
+    private lateinit var testActivity: Activity
+    private lateinit var youTubePlayerView: YouTubePlayerView
+
+    private fun assertFullScreenEntered(fullScreenHelper: FullScreenHelper, youTubePlayerView:YouTubePlayerView) {
```
PierfrancescoSoffritti: Rename to `assertIsFullScreen`.

PierfrancescoSoffritti: Add space here: `youTubePlayerView: YouTubePlayerView`.
Here and below.

core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/FullScreenHelperTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        assertTrue("Testing if isFullScreen boolean is set to true", fullScreenHelper.isFullScreen)
+        assertTrue("Testing if height is MATCH_PARENT", youTubePlayerView.layoutParams.height == ViewGroup.LayoutParams.MATCH_PARENT)
+        assertTrue("Testing if width is MATCH_PARENT", youTubePlayerView.layoutParams.width == ViewGroup.LayoutParams.MATCH_PARENT)
+    }
+
+    private fun assertFullScreenExited(fullScreenHelper: FullScreenHelper, youTubePlayerView:YouTubePlayerView) {
```
PierfrancescoSoffritti: `assertIsNotFullScreen`

core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/PlaybackResumerTest.kt
```diff
@@ -0,0 +1,61 @@
...
+ * Note: The state combinations tested are:
+ *  lifecycle Resumed/Paused -> All PlayerErrors -> All PlayerState -> Random second -> Random videoId
+ */
+@LargeTest
+class PlaybackResumerTest {
+
+    private val isLifeCycleResumedStates = arrayOf(true, false)
+
+    @Test
+    fun testResumeCallsAppropriateYouTubePlayerMethodUsingVariousStates() {
```
PierfrancescoSoffritti: There is way too much going on in this tests, I struggle to understand what's happening. I'd suggest splitting it into multiple tests.

In general, when testing, is good practice to avoid writing tests that are too generic. It's better to write multiple tests and test every case "manually" instead of writing loops that test everything etc.

core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/YouTubePlayerTrackerTest.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.YouTubePlayer
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.YouTubePlayerTracker
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.mockito.Mockito.mock
+import kotlin.random.Random
+
+/**
+ * This class tests if YouTubePlayerTracker updates its attributes correctly
+ *
```
PierfrancescoSoffritti: Remove line.

core/src/androidTest/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/YouTubePlayerTrackerTest.kt
```diff
@@ -0,0 +1,75 @@
...
+ */
+@LargeTest
+class YouTubePlayerTrackerTest {
```
PierfrancescoSoffritti: These tests are not very useful, you are setting things and then asserting that what you just set is true.

A better way to test this would be to attach it to a real YouTubePlayer, interact with the player and then assert that things are as expected.

AbelTesfaye: I agree these tests are a bit pointless, so I'm planning on making the rest of the method tests like the following, let me know if there is a better way?

```kotlin
    /**
     * Observes the player for a period of time (observationTimeInMs) , then checks if
     * [YouTubePlayerTracker] has updated its state accordingly
     */
    @Test
    fun testOnStateChange() {
        // Prepare
        val observationTimeInMs = 20*1000L
        val youtubePlayerTracker = YouTubePlayerTracker()

// [1] The assert listener should be added *after* youtubePlayerTracker, but this currently works since HashSet doesn't insert items in order
        youTubePlayerView.addYouTubePlayerListener(object : AbstractYouTubePlayerListener(){
            override fun onStateChange(youTubePlayer: YouTubePlayer, state: PlayerConstants.PlayerState) {
                // Assert
                assertTrue(state == youtubePlayerTracker.state)
            }
        })

        youTubePlayerView.addYouTubePlayerListener(youtubePlayerTracker)

        Thread.sleep(observationTimeInMs)
    }

```

The downsides of this approach:
1. Since everything will be manual, the code will be super long (but will rigorously test the functionality) 
2. We might need to change the youtubePlayer listener container to some thing more predicable/with order (look at comment "[1]",  it currently uses [HashSet](https://github.com/PierfrancescoSoffritti/android-youtube-player/blob/856e9aafbb66d3e206fae50502502d2fd9921c7f/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt#L29), maybe use array or linked list?)
3. Test might take longer to finish since the player waits for 20 seconds for each test

So, this is what I'm proposing, should I proceed with this approach?

PierfrancescoSoffritti: This makes sense, it could be an interesting exercise but to be fair it seems like a lot of effort to test a class that does essentially nothing.

I'd say we can skip this for now, unless you want to play more with tricky tests.

AbelTesfaye: :) Well, I guess its better to skip this class since it would be implicitly tested when I make the view tests later.

Also `PlaybackResumer` is very similar to this class, so I think I should skip that too. 

So `FullScreenHelperTest` will be the only test in this PR, i'll remove the other tests.

Issue #459, opened 2019-10-22
Memory not being released on destroying activity


divyas93:
I have observed that when activity is destroyed then all the instances of the library code are not finished and thus the memory taken by library code isn't released.

Could you please look into it.

PierfrancescoSoffritti:
Hey, how do you know this is happening? 
Feel free to send a pull request.

acvt007:
I use your library as an item in my RecyclerView in an Activity.
Below is my LeakCanary info:

ApplicationLeak(className=com.abc.ArticleDetailActivity, leakTrace=
┬
├─ org.chromium.device.power_save_blocker.PowerSaveBlocker
│    Leaking: NO (a class is never leaking)
│    GC Root: Global variable in native code
│    ↓ static PowerSaveBlocker.b
│                              ~
├─ java.util.HashMap
│    Leaking: UNKNOWN
│    ↓ HashMap.table
│              ~~~~~
├─ java.util.HashMap$Node[]
│    Leaking: UNKNOWN
│    ↓ array HashMap$Node[].[0]
│                           ~~~
├─ java.util.HashMap$Node
│    Leaking: UNKNOWN
│    ↓ HashMap$Node.key
│                   ~~~
├─ com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.WebViewYouTubePlayer
│    Leaking: YES (View.mContext references a destroyed activity)
│    mContext instance of com.abc.ArticleDetailActivity with mDestroyed = true
│    View#mParent is null
│    View#mAttachInfo is null (view detached)
│    View.mWindowAttachCount = 2
│    ↓ WebViewYouTubePlayer.mContext
╰→ com.theepochtimes.news.features.articledetails.ArticleDetailActivity
​     Leaking: YES (WebViewYouTubePlayer↑ is leaking and Activity#mDestroyed is true and ObjectWatcher was watching this)
​     key = 4ee2961d-1584-4a7a-8eab-d2c91687a56a
​     watchDurationMillis = 6575
​     retainedDurationMillis = 1567
, retainedHeapByteSize=295071)

IvanLu101:
Hello, 
This problem can be reproduced in Sample App.
After adding debugImplementation 'com.squareup.leakcanary:leakcanary-android:2.0-beta-3' into build.gradle of core-sample-app, launch "ayp API" and click "Fragment Example". Then press back key to go back to main page. You will see leakcanary reports the leak(refer to the attachment).
[...]

PR #566, opened 2020-05-18
Fix for issue #459

aalloul:
When onStop is called, LegacyYoutubePlayer doesn't unregister the networkListener and this seems to cause a memory leak. I added the same code as in line 181 of LegacyYoutubePlayer which seems to do the trick.


PierfrancescoSoffritti:
Hey, thanks for the PR. Shouldn't we register the network receiver again in `onStart`?

aalloul:
Sorry for the late reply, been busy with other stuff... you're probably right but then maybe onResume?

PR #592, opened 2020-06-29
Adding allowSeekAhead to seekTo method

rwrz:
- this allows you to seek ahead, even if the video isn't buffered yet (true as default option, so no breakingChanges, and the seekAhead is the normal expected behavior)
- tested on core and chromecast (both worked)
- removed .idea project files from git


rwrz:
Meh. Just realize the allowSeekAhead is already being made on the "seekTo" JS side. Sorry about it.

PR #647, opened 2020-10-09
WIP: Add playback rate implementation

### LINKED ISSUE #101: Add method to change playback speed

### LINKED ISSUE #645: Is it possible to slow down or speed up the videos?
I noticed there is rate in the enum

 enum class PlaybackRate {
        UNKNOWN, RATE_0_25, RATE_0_5, RATE_1, RATE_1_5, RATE_2
    }

Is there a recommended way to set the playbackRate? (Sorry I am not a kotlin guy so I find it a bit hard to navigate in that code!)

Thanks!

malliaridis:
This commit adds playback rate support. It also removes the enum
of playback rates from the constants, since the playback rate can
be any value between `0.25` and `2.0`.

TODO: Check if the chromecast command "SET_PLAYBACK_RATE" is correct and works (I have no idea how to test for chromecast).

Tested on API 30 and works like a charm (except the limitations of the value set by the YouTube API).

Example usage of the new feature: 
```java

@Override
protected void onCreate(Bundle savedInstanceState) {
    super.onCreate(savedInstanceState);
    setContentView(R.layout.activity_simple_example);

    YouTubePlayerView youTubePlayerView = findViewById(R.id.youtube_player_view);

    youTubePlayerView.addYouTubePlayerListener(new AbstractYouTubePlayerListener() {
        @Override
        public void onPlaybackRateChange(YouTubePlayer youTubePlayer, float playbackRate) {
            super.onPlaybackRateChange(youTubePlayer, playbackRate);
            Log.d("PLAYBACK", "Playback changed to:" + playbackRate);
        }
    });

    youTubePlayerView.getYouTubePlayerWhenReady(youTubePlayer -> youTubePlayer.setPlaybackRate(1.5f));

    getLifecycle().addObserver(youTubePlayerView);
}
```

Closes #101.
Closes #645.

### DIFF
--- a/chromecast-receiver/js/YouTubePlayer.js
+++ b/chromecast-receiver/js/YouTubePlayer.js
@@ -95,7 +95,7 @@ function YouTubePlayer(communicationConstants, communicationChannel) {
     }
 
     // JAVA to WEB functions
-    function seekTo(startSeconds) {        	
+    function seekTo(startSeconds) {
         player.seekTo(startSeconds, true)
     }
 
@@ -133,11 +133,15 @@ function YouTubePlayer(communicationConstants, communicationChannel) {
         player.setVolume(volumePercent)
     }
 
+    function setPlaybackRate(suggestedRate) {
+        player.setPlaybackRate(suggestedRate)
+    }
+
     function getActions() {
         return actions
     }
 
-    const actions = { seekTo, pauseVideo, playVideo, loadVideo, cueVideo, mute, unMute, setVolume }
+    const actions = { seekTo, pauseVideo, playVideo, loadVideo, cueVideo, mute, unMute, setVolume, setPlaybackRate }
     
     return {
         initialize,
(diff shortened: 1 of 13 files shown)


malliaridis:
Can somebody test this on chromecast to see if the command is correct? Or provide at least information where I can find these commands to check it myself. This would complete this PR and remove it from WIP state.

PierfrancescoSoffritti:
> The YouTubePlayer does not guarantee that the playback rate will actually change or change to the specified value.

what does this mean?

It is probably not going to work on chromecast, we need to add support for `ChromecastCommunicationConstants.SET_PLAYBACK_RATE` in the javascript code (chromecast-receiver)

malliaridis:
From the IFrame API under [`player.setPlaybackRate(suggestedRate:Number):Void`](https://developers.google.com/youtube/iframe_api_reference#setPlaybackRate): "_[...] Calling this function does not guarantee that the playback rate will actually change. However, if the playback rate does change, the onPlaybackRateChange event will fire, and your code should respond to the event rather than the fact that it called the setPlaybackRate function._" 

I guess they mean that there are specific situations, where `setPlaybackRate` is ignored. Might be for example when you have set the max speed (`2.0`) and call `setPlaybackRate(3.0)` for example. Might also be the case if the player is in a state that cannot handle the call (I have however no idea how exactly they mean it). 

And about the constant, you are right. I have overseen it because Android Studio hid it from the overview. I hope I haven't forget anything else now.

PierfrancescoSoffritti:
Sorry for never getting to this. Changing playback speed has been implemented in version 11.0.0.

PR #696, opened 2021-01-29
Disable Jetifier

anjalsaneen:
## What does this PR do?
- Disable Jetifier. 

## Why is this PR needed?
- Disabling Jetifier which potentially increase compilation time required to process dependencies.

## Links
- https://adambennett.dev/2020/08/disabling-jetifier/

## Breaks
- [ ] it breaks sample apps implementation since it internally uses `com.psoffritti.librarysampleapptemplate` which needed jetifier. This [plugin](https://github.com/dipien/bye-bye-jetifier) helps to fix and verify those.


PierfrancescoSoffritti:
So you just deleted the sample apps?
I am not going to merge this change. Might fix it myself over the weekend if I have time.

PR #743, opened 2021-05-31
Playlist functionality in the example activity

almahirpm:
With this feature the buttons in the player can play the next and previous video in the playlist.


PierfrancescoSoffritti:
Hi, thanks for the PR. Why do you think this change to the sample app is going to be useful?

almahirpm:
> Hi, thanks for the PR. Why do you think this change to the sample app is going to be useful?

Hello!
I just did a comparison with the Youtube app and this was the only pending thing, this is awsome btw and if you think it´s not necessary i can close this, no problem!!!

PR #790, opened 2021-11-04
ConnectivityManger leaks the context on Android 12

frett:
the context object getSystemService() is called on is leaked via `ConnectivityManager.sInstance`

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/Utils.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/Utils.kt
@@ -8,7 +8,7 @@ import java.io.InputStreamReader
 
 internal object Utils {
     fun isOnline(context: Context): Boolean {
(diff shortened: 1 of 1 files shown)


PierfrancescoSoffritti:
why isn't your patch different from the previous implementation?

frett:
Screen Shot 2021-11-05 at 9 22 48 AM

Here is a debugger stack trace of when the `ConnectivityManager.sInstance` variable is being set with a `ConnectivityManager` instance that references an Activity context. The context object that is passed into the `NetworkListener.onReceive()` method is the activity context class, and not the application context class. So, getting the `applicationContext` out of the Activity before getting the ConnectivityManager service will prevent the Activity leaking via the `sInstance` static variable.

This stack trace is from a Pixel 4a running Android 12, I haven't tried running it on older versions of Android yet to determine if the leak was present before or if it's a new leak in Android 12

PierfrancescoSoffritti:
Hi @frett , I am sorry for the very very late reply. As the single maintainer of the lib it's sometimes hard to keep up with everything going on.

Your PR is correct, the `Context` was indeed leaked. I have fixed this recently, see #924. Again sorry for not addressing this sooner.

PR #840, opened 2022-04-17
Update TimeUtilities.kt

JosiYosi:
Video playing elapsed time and video total duration was formatted only in minutes and seconds. It can be formatted in hours, minutes and seconds for an hours long videos.

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/ui/utils/TimeUtilities.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/ui/utils/TimeUtilities.kt
@@ -10,8 +10,27 @@ object TimeUtilities {
     @SuppressLint("DefaultLocale")
     @JvmStatic
     fun formatTime(timeInSeconds: Float): String {
-        val minutes = (timeInSeconds / 60).toInt()
-        val seconds = (timeInSeconds % 60).toInt()
-        return String.format("%d:%02d", minutes, seconds)
+        var timeString =""
+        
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/ui/utils/TimeUtilities.kt
```diff
@@ -10,8 +10,27 @@ object TimeUtilities {
...
-        return String.format("%d:%02d", minutes, seconds)
+        var timeString =""
+        
+        var hours = timeInSeconds / 3600
+        var minutes = (timeInSeconds % 3600) / 60
+        var seconds = timeInSeconds % 60
+       
+        
+         //formatting hours minutes and seconds in string
+        if(hours==0L && minutes==0L){
```
PierfrancescoSoffritti, line 21: this series of if else statement looks a bit convoluted.

PierfrancescoSoffritti:
Do you know how YouTube handles videos of long duration? 
Also, with version 11 web ui is the default, and recommended way to use the library. I'm not convinced this change is necessary.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/ui/utils/TimeUtilities.kt
JosiYosi: I don't know how it becomes confusing for you the if else statement. It is straight forward I guess. Also, you can use a Kotlin version when key word as well if you want to avoid long lines of codes.

PR #842, opened 2022-04-20
Update DefaultPlayerUiController.kt

JosiYosi:
Adding an implemented method for progress bar . . . . Since a default progress bar cannot be disabled or hidden, a custom progress bar in a controlsContainer view can be hidden by implementing an overridden method. Also, making sure that method is declared in that interface class PlayerUiController


JosiYosi:
Both the default progress bar from YouTube and a customized progress bar within controlContainer view are spinning at the same time, and users may either get confused or may think something is behind going on. It would be better if a progress bar in a controlContainer should have been hidden.

PierfrancescoSoffritti:
Hi, starting from version 11 this shouldn't not be an issue anymore. Since the default UI is webUi. So there is only one progress bar.

JosiYosi:
Ok thanks, when you say version 11 do you mean android version 11 or your repository version 11 though? I have seen your latest repo version is 11.

PierfrancescoSoffritti:
Yes I meant version 11 of the library.

PR #844, opened 2022-04-20
Update PlayerUiController.kt

JosiYosi:
Adding an interface method showProgressBar and declare it


PierfrancescoSoffritti:
Hi, what is the purpose of this method? and where is the implementation?

PierfrancescoSoffritti:
I just saw the other PR. I don't think they should be separate.

JosiYosi:
The purpose of adding that method was to disable a customized progress bar since the default one cannot be disabled. But in your other comment you have said that no need to worry about it anymore from version 11 bc the default player is webUi. If that is the case, no need to add the suggested method then.

PierfrancescoSoffritti:
See #842

PR #849, opened 2022-04-29
Hide caption fix

Serkali-sudo:
So, for some reason ccLoadPolicy doesnt work anymore so i found an alternative way to disable captions.
adding `player.unloadModule('captions');`  to `ayp_youtube_player.html` does the trick.

Heres a basic example that shows you how to disable captions:
```java
public class CaptionDisableExample extends AppCompatActivity {
    
    private YouTubePlayerView youTubePlayerView;

    @Override
    protected void onCreate(@Nullable Bundle savedInstanceState) {
        super.onCreate(savedInstanceState);
        youTubePlayerView = findViewById(R.id.youtube_player_view);

        getLifecycle().addObserver(youTubePlayerView);
        youTubePlayerView.addYouTubePlayerListener(new AbstractYouTubePlayerListener() {
            @Override
            public void onReady(@NonNull YouTubePlayer youTubePlayer) {
                String videoId = "yS1ibDImAYU";
                youTubePlayer.loadVideo(videoId, 0);
            }

            @Override
            public void onStateChange(@NonNull YouTubePlayer youTubePlayer, @NonNull PlayerConstants.PlayerState state) {
                if (state == PlayerConstants.PlayerState.PLAYING) {
                    youTubePlayer.killCC();
                }

            }


        });

    }
}
```

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
@@ -25,6 +25,7 @@ interface YouTubePlayer {
 
     fun mute()
     fun unMute()
+    fun killCC()
 
     /**
      * @param volumePercent Integer between 0 and 100
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
@@ -65,6 +65,10 @@ internal class WebViewYouTubePlayer constructor(context: Context, attrs: Attribu
     override fun unMute() {
         mainThreadHandler.post { loadUrl("javascript:unMute()") }
     }
+    
+    override fun killCC(){
+        mainThreadHandler.post{ loadUrl("javascript:killCC()")}
+    }
 
     override fun setVolume(volumePercent: Int) {
         require(!(volumePercent < 0 || volumePercent > 100)) { "Volume must be between 0 and 100" }
--- a/core/src/main/res/raw/ayp_youtube_player.html
+++ b/core/src/main/res/raw/ayp_youtube_player.html
@@ -139,6 +139,10 @@
         function unMute() {
             player.unMute();
         }
+            
+        function killCC(){
+            player.unloadModule('captions');
+        }    
 
         function setVolume(volumePercent) {
             player.setVolume(volumePercent);


core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
```diff
@@ -25,6 +25,7 @@ interface YouTubePlayer {
 
     fun mute()
     fun unMute()
+    fun killCC()
```
PierfrancescoSoffritti, line 28: this should be called `unloadCaptions`

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
```diff
@@ -65,6 +65,10 @@ internal class WebViewYouTubePlayer constructor(context: Context, attrs: Attribu
     override fun unMute() {
         mainThreadHandler.post { loadUrl("javascript:unMute()") }
     }
+    
+    override fun killCC(){
+        mainThreadHandler.post{ loadUrl("javascript:killCC()")}
+    }
```
PierfrancescoSoffritti, line 71: add a space before the last curly brace

PierfrancescoSoffritti:
Hi, what issue are you trying to solve? every video plays with captions? can you show me an example of the problem?

Serkali-sudo:
> Hi, what issue are you trying to solve? every video plays with captions? can you show me an example of the problem?

well current way to disable caption is using ccLoadPolicy but it doesnt work anymore.so i recommended alternative way

PierfrancescoSoffritti:
Can you send me a video where the issue happens?

Serkali-sudo:
> Can you send me a video where the issue happens?

i dont think we need a video.my problem is i cant disable captions with current method.

PierfrancescoSoffritti:
Can you send me a video where the captions show up? so i can test what's going on?

PierfrancescoSoffritti (commented):
Can you send the pull request to the dev branch instead of master?

Serkali-sudo:
@PierfrancescoSoffritti Done!

PierfrancescoSoffritti:
Can you send me a video where the captions show up? so i can test what's going on?

Serkali-sudo:
@PierfrancescoSoffritti 
So as you see in the video every video i open starts with captions even though i set ccLoadPolicy = 1 to disable captions.Like this:
```java
IFramePlayerOptions build = new IFramePlayerOptions.Builder().controls(0).rel(0).ccLoadPolicy(1).build();
```


But apperantly setting ccLoadPolicy to 1 doesn't disable captions anymore. Because it still shows captions. But my pull request fixes it

https://user-images.githubusercontent.com/59535990/166449902-e29ff321-74dd-4f86-b23a-495e27d8c6c6.mp4

PierfrancescoSoffritti:
ok, I understand. Thanks!

I will try everything out in the next few days (over the weekend at the latest) and merge the PR then :)

PierfrancescoSoffritti:
I'm looking at the [IFrame player API](https://developers.google.com/youtube/iframe_api_reference), and it seems like the `unloadModule` method is not documented there. Which makes me think it's not a public API. Which if it's the case is concerning, since it means YouTube can break our library at any time.

How did you find out about the existence of the method?

Serkali-sudo:
@PierfrancescoSoffritti 
You are right i didnt found this method from docs.
`unloadModule` is basically javascript object method that allows you to unload spesific js files.
For example `player.unloadModule('captions')` disables `captions.js` file.
(i mean whatever you type inside unloadModule it will disable that js file)
So if disabling js files without using docs methods is breaking Youtube terms then sorry for wasting your time.

PierfrancescoSoffritti:
The problem are not YouTube terms of service. The problem is that it's not good practice for us to rely on this, as it's not a public API and it would break our lib if YouTube changes something. For example if they rename the module to something else (eg: "captions" or "cc").

Here is my suggestion:
Let's replace the method in the interface, from `unloadCaptions()` to `unloadModule(module: String)`

But first I would like to see the documentation of the Javascript `unloadModule` API. Could you add the link to this discussion?

Issue #348, opened 2019-04-10
Add support for playing playlists


JoseMarioPerezCotes:
hello pierfrancesco, I want know if is posible to play a playlist just like the official youtube api?

PierfrancescoSoffritti:
Hi, currently the only way to play a playlist is to play new videos each time one finishes.

A future improvement for the library could be implementing the [playlist](https://developers.google.com/youtube/player_parameters#playlist) parameter in [IFramePlayerOptions](https://github.com/PierfrancescoSoffritti/android-youtube-player#iframeplayeroptions).

acrolink:
That would be a great enhancement.

mumenAlmadaineh:
Was this feature added? -If yes, how can I add a playlist to API (Android Studio)

PierfrancescoSoffritti:
There is no support for playlists.

mumenAlmadaineh:
😥

PierfrancescoSoffritti:
thanks to @flauschtrud playlists are now supported in the library :)

PR #863, opened 2022-07-04
Added support for playlists

flauschtrud:
I added support for playlists as described in https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/348 by implementing the list and listType parameters for the IFramePlayerOptions. If a video is chosen in the playlist any YouTubePlayerListener will be notified in the onVideoId callback.

I also added an example to demonstrate the usage.

I also thought about adding a playlist attribute to the YouTubePlayerView, but didn't want to change the API without prior consultation.

Let me know if you like these changes.

And thank you very much for your work, I'm very happy to use it in my project.

### DIFF
--- a/core-sample-app/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/sampleapp/examples/playlistExample/PlaylistExampleActivity.java
+++ b/core-sample-app/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/sampleapp/examples/playlistExample/PlaylistExampleActivity.java
@@ -1,0 +1,57 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.sampleapp.examples.playlistExample;
+
+import android.content.res.Configuration;
+import android.os.Bundle;
+
+import androidx.annotation.NonNull;
+import androidx.appcompat.app.AppCompatActivity;
+
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.YouTubePlayer;
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.AbstractYouTubePlayerListener;
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.options.IFramePlayerOptions;
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.YouTubePlayerUtils;
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView;
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.sampleapp.utils.VideoIdsProvider;
+import com.pierfrancescosoffritti.aytplayersample.R;
+
+public class PlaylistExampleActivity extends AppCompatActivity {
+
+    private YouTubePlayerView youTubePlayerView;
+    private static final String PLAYLIST_ID = "PLEpEmEcrrKJUhZkyIAgQ17Oxyd3fx_y1j";
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_playlist_example);
+
+        youTubePlayerView = findViewById(R.id.youtube_player_view);
+        getLifecycle().addObserver(youTubePlayerView);
+
+        initYouTubePlayerView();
+    }
+
+    private void initYouTubePlayerView() {
+        IFramePlayerOptions iFramePlayerOptions = new IFramePlayerOptions.Builder()
+                .controls(1)
+                .listType("playlist")
+                .list(PLAYLIST_ID)
+                .build();
+
+        getLifecycle().addObserver(youTubePlayerView);
+
+        youTubePlayerView.initialize(new AbstractYouTubePlayerListener() {}, true, iFramePlayerOptions);
+    }
+
+    @Override
+    public void onConfigurationChanged(@NonNull Configuration newConfig) {
+        super.onConfigurationChanged(newConfig);
+
+        // Checks the orientation of the screen
+        if (newConfig.orientation == Configuration.ORIENTATION_LANDSCAPE) {
+            youTubePlayerView.enterFullScreen();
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
```diff
@@ -34,6 +34,8 @@ class IFramePlayerOptions private constructor(private val playerOptions: JSONObj
             private const val MODEST_BRANDING = "modestbranding"
             private const val CC_LOAD_POLICY = "cc_load_policy"
             private const val CC_LANG_PREF = "cc_lang_pref"
+            private const val LIST = "list"
```
PierfrancescoSoffritti, line 37: can you explain what "list" and "listType" do?

core-sample-app/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/sampleapp/examples/playlistExample/PlaylistExampleActivity.java
```diff
@@ -0,0 +1,57 @@
...
+        youTubePlayerView = findViewById(R.id.youtube_player_view);
+        getLifecycle().addObserver(youTubePlayerView);
+
+        initYouTubePlayerView();
+    }
+
+    private void initYouTubePlayerView() {
+        IFramePlayerOptions iFramePlayerOptions = new IFramePlayerOptions.Builder()
+                .controls(1)
+                .listType("playlist")
```
PierfrancescoSoffritti, line 36: why are we passing "playlist" here? what does it mean?

core/src/main/res/raw/ayp_youtube_player.html
```diff
@@ -96,6 +99,15 @@
                 YouTubePlayerBridge.sendVideoDuration(videoDuration);
             }
 
+            function sendCurrentVideoIdFromPlaylist(player) {
+                var playlist = player.getPlaylist();
+                if ( typeof playlist !== 'undefined' && Array.isArray(playlist) && playlist.length > 0 ) {
+                    var index = player.getPlaylistIndex();
```
PierfrancescoSoffritti, line 105: are we 100% sure this is always going to be an integer? should we check if it is? can you please send a link to the doc?

core/src/main/res/raw/ayp_youtube_player.html
```diff
@@ -75,6 +75,9 @@
                     
                     startSendCurrentTimeInterval();
                     sendVideoData(player);
+
+                    sendCurrentVideoIdFromPlaylist(player);
```
PierfrancescoSoffritti, line 79: i wonder if a better place for this would be `YT.PlayerState.BUFFERING`. What do you think?

PierfrancescoSoffritti (commented):
Thanks for the pull request. I added a few comments. In general I would like to have links to the original doc to better understand how things will work.

But overall the pull request makes sense.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
flauschtrud: These are parameters of the YouTube player as described here: https://developers.google.com/youtube/player_parameters#Selecting_Content_to_Play
"listType" can either be "playlist" or "user_uploads", although "user_uploads" is kind of deprecated because it requires a YouTube username which only quite old channels seem to have.
"list" is either a playlist id (if "listtype" is set to "playlist") or the username (if "listtype" is set to "user_uploads").
Both ways work with this implementation.

core-sample-app/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/sampleapp/examples/playlistExample/PlaylistExampleActivity.java
flauschtrud: As described above "playlist" is one of the two valid options for the "listtype" parameter (the other being "user_uploads").

core/src/main/res/raw/ayp_youtube_player.html
flauschtrud: Hmm, both states (BUFFERING and PLAYING) have the disadvantage that they are are potentially entered multiple times while sending the same video id each time (for example if the user uses the seekbar to jump ahead). Although I assume that the BUFFERING state is reached first, which probably makes it the better place.

CUED does not work, since it seems to only be reached when a video (or playlist) is cued directly via a player function (`cueVideoById` or `cuePlaylist`) and not by interacting with the embedded playlist.

I can change the location to BUFFERING unless you have you an idea on how to prevent sending redundant video ids.

PierfrancescoSoffritti: > Hmm, both states (BUFFERING and PLAYING) have the disadvantage that they are are potentially entered multiple times while sending the same video id each time (for example if the user uses the seekbar to jump ahead). Although I assume that the BUFFERING state is reached first, which probably makes it the better place.

I didn't consider this. I'd say this has to be solved before we can merge this pull request. We don't want to send the video id each time the video is BUFFERING or PLAYING. As this is not the contract for the listener.

flauschtrud: I agree. I just looked into it again and think that the UNSTARTED event is fine. The doc says "When the player first loads a video, it will broadcast an unstarted (-1) event." I must have missed this before. I just tested it with the sample app and it works well: we only get a videoid once per video. Should I commit this change?

PierfrancescoSoffritti: Great news, thanks for checking! I will take a final look over the weekend :)

PierfrancescoSoffritti (approved):
I have added a few comments for documentation, thanks a lot for your contributions!

Issue #830, opened 2022-03-24
onReady never called when enter and exit activity quickly


ducdat-ptit:
https://user-images.githubusercontent.com/12276929/159845665-beb807da-adb2-412d-8adb-0622808b264f.mp4

PierfrancescoSoffritti:
did you make any change to the code?

ducdat-ptit:
> did you make any change to the code?

No. I didn't. The video is from sample apk:  ([core v11.0.1.apk](https://github.com/PierfrancescoSoffritti/android-youtube-player/blob/master/core-sample-app/apk/core%20v11.0.1.apk)

nerooke:
> did you make any change to the code?

 I have same issue, player do not ready in this case.

PierfrancescoSoffritti:
Which device are you using?

ducdat-ptit:
I checked on Samsung s10 (api 31), Samsung note10 (api 31), Pixel 4 (api 30), Xiaomi redmi note 8 (api 30)

PierfrancescoSoffritti:
I just tried and everything works fine for me. Maybe it's something local to your build?

ralexey:
@PierfrancescoSoffritti 
I tried to use your library at the project for my client and i got the same behaviour.
Tested on several devices with Android 12 (Pixel 3a XL in my case)
I can open/close 2-3 time the activity and it shows the YT video, but it shows the black screen on the next time and doesn't call the onReady. But if i close/open the activity again - it works fine.

I used the manual initialisation, set the lifecycle and manual load or cue the YT video id.

Wayne-Wang-Wayne:
I got the same issue here, what is the fix QQ?

Fangzsx:
Faced the same issue. Any workaround to fix this?

minseokchoi:
I have same issue
chrome app 98.0.4758.101 No problem
chrome app 100.0.4896.88 there is a problem

This is a problem caused by the latest Chrome update.
If your app is large, you may have to wait longer. 5 to 10 seconds

@PierfrancescoSoffritti
Any workaround to fix this?

Fangzsx:
Hi guys. I know this seems messy but I tried overriding the onError. I called the youtubePlayer.loadVideo again and insert the responding videoID. If the youtube player display playback error, if you click the video, loadvideo will be called once again hence, loading the video again.

brain-deaf:
[...]

Issue #869, opened 2022-07-25
Why are we not using the WebView's cache?


flauschtrud:
I noticed that caching is explicitly set to `WebSettings.LOAD_NO_CACHE` in WebViewYouTubePlayer.initWebView.

Is there a reasoning behind this?

I set it to `WebSettings.LOAD_DEFAULT` and did not encounter any problems, while saving a lot of network traffic (thumbnails, JS and CSS for the embedded player).

The doc for reference: https://developer.android.com/reference/android/webkit/WebSettings#LOAD_NO_CACHE
`WebSettings.LOAD_NO_CACHE`: Don't use the cache, load from the network.
`WebSettings.LOAD_DEFAULT`: Default cache usage mode. If the navigation type doesn't impose any specific behavior, use cached resources when they are available and not expired, otherwise load resources from the network.

PierfrancescoSoffritti:
Yeah, we should probably use the cache.

PierfrancescoSoffritti:
Feel free to send a PR for this, if you have time. Otherwise I'll do it before releasing the next version.

PierfrancescoSoffritti:
Thanks to @flauschtrud , this will be available starting from version `11.1.0`.

PR #873, opened 2022-08-03
Potential fix for issue #830

flauschtrud:
I believe to have found a fix for https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/830.

I changed the location where the IFrame Player API code is loaded to do it asynchronously (see the [doc](https://developers.google.com/youtube/iframe_api_reference#Getting_Started))
Apparently there are race conditions otherwise where it is possible that `onYouTubeIframeAPIReady` is called before the library's code is loaded properly.

Now I can neither reproduce the issue in my own project nor in the example project (the scenario that the OP showed in the first post's video which I could reproduce before).

### DIFF
--- a/core/src/main/res/raw/ayp_youtube_player.html
+++ b/core/src/main/res/raw/ayp_youtube_player.html
@@ -16,7 +16,7 @@
         <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
         <!-- defer forces the library to execute after the html page is fully parsed. -->
         <!-- This is needed to avoid race conditions, where the library executes and calls `onYouTubeIframeAPIReady` before the page is fully parsed. -->
-        <!-- See ##873 on GitHub -->
+        <!-- See #873 on GitHub -->
         <script defer src="https://www.youtube.com/iframe_api"></script>
     </head>
 
--- a/core/src/main/res/raw/ayp_youtube_player.html
+++ b/core/src/main/res/raw/ayp_youtube_player.html
@@ -21,6 +21,7 @@
     </body>
 
     <script type="text/javascript">
+        loadIFramePlayerAPI()
         
         var UNSTARTED = "UNSTARTED";
         var ENDED = "ENDED";
@@ -34,13 +35,6 @@
 
         var timerId;
 
-        // load the IFrame Player API code asynchronously.
-        var tag = document.createElement('script');
-        tag.src = "https://www.youtube.com/iframe_api";
-        var firstScriptTag = document.getElementsByTagName('script')[0];
-        firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
-
-        // creates an <iframe> (and YouTube player) after the API code downloads.
     	function onYouTubeIframeAPIReady() {
 
             YouTubePlayerBridge.sendYouTubeIFrameAPIReady();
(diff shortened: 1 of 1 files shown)


PierfrancescoSoffritti:
Ah interesting! that could make the difference.

Sorry to keep you waiting on the other pull request, i've been quite busy lately. I'll make sure to go over this PR and the other one by the end of this week.

PierfrancescoSoffritti:
I tested your change, it does to fix the bug! thanks a lot for looking into it!

I made a small change to your PR, to make the code a bit more readable.

Let me know if it looks good to you :)

flauschtrud:
Huh, I'm a little confused now about Google's doc since they are so explicit about doing things asynchronously although I'd agree with you that this would not be the smartest solution. I mean, it's obviously better than a blocking fetch and execute in the head, but it still seems to rely on the API code being big enough in comparison to the document being parsed (which probably works but still seems risky).

I just updated the PR to use the `defer` attribute. This should basically be the same as the changes that you made, but the most minimal. So we get an asynchronous download but the script will only be executed after the HTML has been parsed (I tested again with my project and the example app).

BTW, will the [cache usage](https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/869) be added to the next release too? I didn't create a PR for this one since it seemed a little overkill for just a parameter.

PierfrancescoSoffritti:
> I'm a little confused now about Google's doc since they are so explicit about doing things asynchronously

Can you point me to that part of the doc?

I think it would be useful to clarify why the bug is happening in the first place, and why this solution fixes it.

The documentation for `defer` says:

> If defer is present (and not async): The script is downloaded in parallel to parsing the page, and executed after the page has finished parsing

In the case of the IFrame API it's not clear to me what does it mean that the scrip is executed. Downloading the YouTube library just makes it available for our local Javascript. Nothing is executed until our script runs and invokes the IFrame API.

Is the bug happening because our script runs before the IFrame API is available? Or is it happening because everything works fine with the IFrame API, but when functions like `onReady` are called the `YouTubePlayerBridge` is not available yet?

Let me know if you have an idea of what's going on.

flauschtrud:
> Can you point me to that part of the doc?

[Here](https://developers.google.com/youtube/iframe_api_reference#Getting_Started) they are describing how to load the library (which is basically the initial version of my fix) and they are quite explicit about loading it asynchronously (see also comment number 3 of this section of the doc) without acknowledging that the API code still might get executed before the initial document is loaded fully. So I'm confused because I believe your version of the solution or my version with the defer attribute is better i.e. safer than what they are suggesting in the doc.

> In the case of the IFrame API it's not clear to me what does it mean that the scrip is executed. Downloading the YouTube library just makes it available for our local Javascript. Nothing is executed until our script runs and invokes the IFrame API.

That's not true. See the [requirements part of the doc](https://developers.google.com/youtube/iframe_api_reference#Requirements).

It says

> Any web page that uses the IFrame API must also implement the following JavaScript function:

>    onYouTubeIframeAPIReady – The API will call this function when the page has finished downloading the JavaScript for the player API, which enables you to then use the API on your page. Thus, this function might create the player objects that you want to display when the page loads.

So the bug is happening because sometimes the API code is loaded and executed too fast and the library's `onYouTubeIframeAPIReady` can not be called because it has not been parsed fully. And apparently changes in Chrome/Webview's latest versions make this more likely.

It's a really weird API design...

PierfrancescoSoffritti:
This makes sense now, thanks for explaining. I forgot that the library calls `onYouTubeIframeAPIReady` when it's done.

> It's a really weird API design
This seems like a bug in the library, doesn't look like intended behaviour to me.

I will update the comment in the file and merge your PR, thanks a lot for looking into this!

flauschtrud:
> This seems like a bug in the library, doesn't look like intended behaviour to me.

Hmm, since they are documenting it this way I'd assume it was intended (but probably a choice that would not be made if the library was written today). The whole doc seems quite outdated, too (they are referring to Stack Overflow issues from 2009...).

But anyway, I'm really happy that the issue is fixed, that has driven me crazy for a while.

PierfrancescoSoffritti:
Yes, [updates have not been that frequent](https://developers.google.com/youtube/player_parameters#Revision_History).

I will try to push out a new version over the weekend!

PR #926, opened 2023-02-19
Add fullScreen function in IFramePlayerOptions for set fs value

Praveen-Pable:
Added one method `fullScreen`  in `IFramePlayerOptions` class for set value of fs parameter.


PierfrancescoSoffritti:
Hi, thanks for the pull request. Can you explain why this is needed?

Praveen-Pable:
> Hi, thanks for the pull request. Can you explain why this is needed?

Hi, this function need to display full screen button with normal youtube control. Default value is set to 0.

PierfrancescoSoffritti:
So you want to add this function to be able to see the full screen button? as far as i know the button doesn't work on Android. Did you verify that?

Praveen-Pable:
For more reference, please find below we 2 scenario of `fs` affect youtube ui. Observe bottom right corner of image for fullscreen button option.

| with fs=0 | with fs=1 |
|--|--|
|img1-screeshot|img2-screenshot|

Praveen-Pable:
> So you want to add this function to be able to see the full screen button? as far as i know the button doesn't work on Android. Did you verify that?

Yes, you are correct. But still if user want to have that option to display button. 
In `WebViewYouTubePlayer` class, when we set `WebChromeClient`, we have `onShowCustomView` and `onHideCustomView` callback function which helps user to manually handle full screen functionality for youtube video.

PierfrancescoSoffritti:
Interesting, i wasn't aware of [`WebChromeClient#onShowCustomView`](https://developer.android.com/reference/android/webkit/WebChromeClient#onShowCustomView(android.view.View,%20android.webkit.WebChromeClient.CustomViewCallback)). I think it would be really helpful if you could write a new sample activity in the sample app. It would help me and other people understand how to use this.

Praveen-Pable:
> Interesting, i wasn't aware of [`WebChromeClient#onShowCustomView`](https://developer.android.com/reference/android/webkit/WebChromeClient#onShowCustomView(android.view.View,%20android.webkit.WebChromeClient.CustomViewCallback)). I think it would be really helpful if you could write a new sample activity in the sample app. It would help me and other people understand how to use this.

Sure, i will create one sample app. Thanks for quick feedback 🙂

Praveen-Pable:
Hi, i have added sample code for full screen. For navigate to sample full screen example, please find 2 option in menu.
Below are some screenshots of sample full screen example:

| Portrait | Landscape |
|--|--|
|WhatsApp Image 2023-02-19 at 6 00 26 PM|WhatsApp Image 2023-02-19 at 6 00 26 PM (1)|

Even if button functionality is not working automatically as per expectation but user can mask it and manually handle as per need like i did in above sample.

Please take a look, waiting for your feedback. :)

PierfrancescoSoffritti:
By looking at the sample Activity, it seems like you added a button to the xml, which when clicked changes the orientation of the Activity. It doesn't seem to be related to the html full screen button?

Praveen-Pable:
In latest commit, i have handle event for display full screen video by using of state change.

| Portrait | Landscape |
|--|--|
|WhatsApp Image 2023-02-19 at 7 05 04 PM (1)|WhatsApp Image 2023-02-19 at 7 05 04 PM|

Praveen-Pable:
> By looking at the sample Activity, it seems like you added a button to the xml, which when clicked changes the orientation of the Activity. It doesn't seem to be related to the full screen button in the IFrame player?

In latest commit, i remove the button. In first example, i try to showcase masking example. And in second example, i try to showcase state handle example. 

In second example, IFrame player full screen button icon change in landscape mode as per expectation.

Praveen-Pable:
Hi @PierfrancescoSoffritti , waiting for your feedback. Please let me know if above solution and suggestion looks good or not.

Praveen-Pable:
Hi @PierfrancescoSoffritti, please update. Waiting for your response.

PierfrancescoSoffritti:
Hi thanks for the ping, I will try to have look before the end of the week.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/listeners/YouTubePlayerListener.kt
```diff
@@ -14,7 +15,7 @@ interface YouTubePlayerListener {
      * Called every time the state of the player changes. Check [PlayerConstants.PlayerState] to see all the possible states.
      * @param state a state from [PlayerConstants.PlayerState]
      */
-    fun onStateChange(youTubePlayer: YouTubePlayer, state: PlayerConstants.PlayerState)
+    fun onStateChange(youTubePlayer: YouTubePlayer, state: PlayerConstants.PlayerState, view: View? = null)
```
PierfrancescoSoffritti, line 18: I find this confusing, why do we need to add a nullable view?

This is problematic for a few reasons:
1. Is a breaking change. Every user of the library will have to change their implementation of `onStateChange`
2. The API is not clear

Maybe this should be a separate method instead?

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/PlayerConstants.kt
```diff
@@ -3,7 +3,7 @@ package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 class PlayerConstants {
 
     enum class PlayerState {
-        UNKNOWN, UNSTARTED, ENDED, PLAYING, PAUSED, BUFFERING, VIDEO_CUED
+        UNKNOWN, UNSTARTED, ENDED, PLAYING, PAUSED, BUFFERING, VIDEO_CUED, SHOW_CUSTOM_VIEW, HIDE_CUSTOM_VIEW
```
PierfrancescoSoffritti, line 6: These two constants are not related to the state of the player, please move them somewhere else

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/listeners/YouTubePlayerListener.kt
Praveen-Pable: Changes are made as per suggestion:

1. Revert code for `onStateChange` method definition, removed `view` parameter
2. Create separate method for handle callback of `onShowCustomView` and `onHideCustomView`

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/PlayerConstants.kt
Praveen-Pable: Changes are made as per suggestion

1. Remove `SHOW_CUSTOM_VIEW` and `HIDE_CUSTOM_VIEW` from PlayerState
2. Update `SimpleFullscreenExampleActivity` code accordingly.

PierfrancescoSoffritti:
Thanks for updating the change. It is a lot more clear now. Can you please send it to the `dev` branch instead of `master`?

I will play with it over the weekend and maybe make some adjustments. Should be merged by Monday :)

PierfrancescoSoffritti:
I've been thinking about this change. As it is it will make things more complicated, because the library already has logic to handle full screen at the view level (instead of webview level). I will need to make more explicit that the current full screen logic is at the view level and will just make the YouTubePlayerView fill its parent. After that is done we can integrate this change.

So it will take a little longer. I'll keep you posted. In the meantime please send the pull request to the `dev` branch

PierfrancescoSoffritti (approved):
lgtm

PierfrancescoSoffritti:
Merged. Thanks @Praveen-Pable for the help!

Praveen-Pable:
> Merged. Thanks @Praveen-Pable for the help!

Thanks @PierfrancescoSoffritti for guiding me throughout the process and for valuable feedback. 🙂👍

PR #936, opened 2023-03-19
feature/auto-fullscreen-on-orientation

Praveen-Pable:
Implement functionality for toggle full screen programmatically as per requirement.


Praveen-Pable:
Please find below screenshot for reference of fullscreen toggle on device orientation change

| Portrait mode | Landscape mode |
|---|---|
|Screenshot_1679247201|Screenshot_1679247274|

core-sample-app/src/main/AndroidManifest.xml
```diff
@@ -90,6 +90,11 @@
             android:configChanges="orientation|screenSize|keyboardHidden|smallestScreenSize|screenLayout"
             android:hardwareAccelerated="true"
             android:label="@string/fullscreen_example" />
+        <activity
```
PierfrancescoSoffritti, line 93: No need to create a new sample activity. Can we reuse the existing full screen sample activity?

core/src/main/res/raw/ayp_youtube_player.html
```diff
@@ -162,5 +162,9 @@
             player.setPlaybackRate(playbackRate);
         }
 
+        function toggleFullscreen() {
+            player.toggleFullscreen();
```
PierfrancescoSoffritti, line 166: Is this function documented somewhere in the IFrame Player API? how did you find out about it?

PierfrancescoSoffritti (commented):
Thanks for the PR! I have left a few comments. I want to make sure that `player.toggleFullScreen` is a public API in the IFrame Player. We don't want to use internal/private stuff as it might break at anytime if YouTube decide to change the API.

core/src/main/res/raw/ayp_youtube_player.html
PierfrancescoSoffritti: I have done some research, that method shows up when `origin` is set to `www.youtube.com`. Therefore I am unsure about how reliable that API is.

Can we try to use [this dom API](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen) to enter full screen instead?

core-sample-app/src/main/AndroidManifest.xml
Praveen-Pable: Noted 👍

Praveen-Pable: Changes are made as per suggestion, removed `FullscreenOnOrientationExampleActivity` and update `FullscreenExampleActivity` accordingly.

core/src/main/res/raw/ayp_youtube_player.html
Praveen-Pable: Refactored code as per suggestion, use dom api for enter and exit fullscreen mode. Please help to review.

core/src/main/res/raw/ayp_youtube_player.html
```diff
@@ -162,5 +162,13 @@
             player.setPlaybackRate(playbackRate);
         }
 
+        function enterFullscreen() {
+            document.getElementById("youTubePlayerDOM").contentWindow.document.getElementById("movie_player").requestFullscreen();
```
PierfrancescoSoffritti, line 166: We can't use private elements ids like "movie_player" they might change at any time. Unless it's documented in the docs?

I will see if there is a way around this. Let me know if you have ideas.

PierfrancescoSoffritti: I have reverted to `toggleFullscreen` as it is the best option.

PierfrancescoSoffritti:
Thanks for the help @Praveen-Pable :)

PR #977, opened 2023-05-04
Replace api with implementation on core gradle.build

kukadiajayesh:


PierfrancescoSoffritti:
Hi, thanks for the PR. could you explain why this is needed?

PR #979, opened 2023-05-04
Remove dex count

kukadiajayesh:
https://jitpack.io/com/github/mikashboks/android-youtube-player/12.0.3/build.log

Based on log, this gradle plugin causing build compile error on jitpack.io.

```No matching variant of com.getkeepsafe.dexcount:dexcount-gradle-plugin:4.0.0 was found. The consumer was configured to find a runtime of a library compatible with Java 8, packaged as a jar, and its dependencies declared externally, as well as attribute 'org.gradle.plugin.api-version' with value '7.5'```


PierfrancescoSoffritti:
Hi, thanks for the PR. The multidex and api changes are not related to dex count though. Could you remove those changes from this PR?

PR #996, opened 2023-06-06
NetworkObserver SDK checks inconsistency resolved

maharshpatel247:


PierfrancescoSoffritti:
min API for `unregisterNetworkCallback` is L. Why do you want to use N?

maharshpatel247:
Ohh didn't see that! I just checked the `observeNetwork` function and there was a check for N. SO I just added N check for `destroy`. I will go ahead and change that in `observeNetwork` to L

maharshpatel247:
On further investigation, I found that [registerDefaultNetworkCallback](https://developer.android.com/reference/android/net/ConnectivityManager#registerDefaultNetworkCallback(android.net.ConnectivityManager.NetworkCallback)) has been added in N so while observing there is a check for N so while destroying too we need to check for N even though `unregisterNetworkCallback` is available in L

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/NetworkObserver.kt
```diff
@@ -36,7 +36,7 @@ internal class NetworkObserver(private val context: Context) {
 
   /** Stop observing network changes and cleanup */
   fun destroy() {
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
```
PierfrancescoSoffritti, line 39: Please add a comment like

```
// Min API for `unregisterNetworkCallback` is L, but we use `registerDefaultNetworkCallback` only for N and above.
```

PierfrancescoSoffritti:
You're right, thanks for finding this!

maharshpatel247:
@PierfrancescoSoffritti Done

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/NetworkObserver.kt
maharshpatel247: Done

PierfrancescoSoffritti (approved):

PierfrancescoSoffritti:
Thanks a lot for the help :)

PR #1055, opened 2023-09-14
Fix changed YouTubePlayerSeekBar path in README.md

tahakorkem:
In README.md, YouTubePlayerSeekBar path in XML was wrong. Maybe its path was changed in the library, but README file was not. This PR contains this minor correction.


PierfrancescoSoffritti:
Thanks for the help!

Issue #1001, opened 2023-06-13
Please add option to have more PlaybackRate specially greater than 2x


Bhavya-gupta127:
Hi, 
# Feature Request
more PlayBackRate options

### Describe your idea
Currently, maximum rate is 2
youTubePlayer.setPlaybackRate(PlayerConstants.PlaybackRate.RATE_2);

On the web, we can change using document.getElementsByTagName("video")[0].playbackRate = x
and there are apps like youtube revanced that support such speed.

PierfrancescoSoffritti:
Hi, feel free to send a PR to verify that this would actually work.

Bhavya-gupta127:
Sure, I will try to do this

Issue #992, opened 2023-05-31
More options to handle the rate in the method youTube Player.setPlaybackRate()


christiansasig:
# Feature Request

### Describe your idea
Hi Everyone,  thank you very much for your spectacular library. Could you please add this rate (PlayerConstants.PlaybackRate.RATE_0_75 -> 0.75f) to the youTubePlayer.setPlaybackRate() method or add a method where the rate can be sent as a float parameter.
Thank you so much

PR #1073, opened 2023-10-25
Add more playback rates

santiwanti:
The Youtube Web Client supports any value between 0.25 and 2.0 so I'm guessing this API should be able to do so as well. I have added two commits, one that adds just the missing 0.75, 1.25, and 1.75 increments and one that adds a function to allow any float value. We might have to check what happens if 2.5 or 100 is sent and maybe a doc indicating that only values between 0.25 and 2.0 are valid.

This PR addresses:
https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/1001
https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/992

### DIFF
--- a/chromecast-sender/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/chromecast/chromecastsender/ChromecastYouTubePlayer.kt
+++ b/chromecast-sender/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/chromecast/chromecastsender/ChromecastYouTubePlayer.kt
@@ -148,9 +148,13 @@ class ChromecastYouTubePlayer internal constructor(private val chromecastCommuni
   }
 
   override fun setPlaybackRate(playbackRate: PlayerConstants.PlaybackRate) {
+    setPlaybackRate(playbackRate.toFloat())
+  }
+
+  override fun setPlaybackRate(playbackRate: Float) {
     val message = JSONUtils.buildFlatJson(
       "command" to ChromecastCommunicationConstants.SET_PLAYBACK_RATE,
-      "playbackRate" to playbackRate.toFloat().toString()
+      "playbackRate" to playbackRate.toString()
     )
 
     chromecastCommunicationChannel.sendMessage(message)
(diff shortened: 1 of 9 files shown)


PierfrancescoSoffritti:
Hi, thanks for the PR!

Did you try passing a value bigger than 2? what happens?

santiwanti:
I just tested it which lead me to fix a few bugs I had introduced. 

When calling it with a value greater than 2 the playback rate is set to 2. Should we add a `require` making sure the user never sends a value greater than 2 or less than 0.25? Or should we document the behaviour indicating that any value outside the range 0.25 and 2 will be coerced so 0.1 or negative numbers will be 0.25 and 3 will be 2?

santiwanti:
The difference with `fun setPlaybackRate(playbackRate: Float)` is that for values like 0.45 it will return 0.45 instead of `PlayerConstants.PlaybackRate.UNKNOWN` which is what `  fun setPlaybackRate(playbackRate: PlayerConstants.PlaybackRate)
` returns

santiwanti:
@PierfrancescoSoffritti sorry to bother you. Anything else that should be done on this PR?

PierfrancescoSoffritti:
I have removed the option to use a generic float to set the playback rate, as it doesn't seem necessary and introduces too much ambiguity in the API.

But happy to add the new enums, thanks for your contribution :)

santiwanti:
@PierfrancescoSoffritti the point of being able to set the playback rate with an arbitrary float is that you can set the value to 1.2 or any other value between 0.25 and 2.0 which doens't have an enum value. This is also allowed in the youtube website if you click on playbackrate and select custom.

PierfrancescoSoffritti:
I understand, why is using the predefined values not enough for your use case?

santiwanti:
In my use case we have to allow any value between 0.25 and 2.0 with increments of 0.01. This is for an app for musicians which might want to play the video at arbitrary speeds. I understand if you don't want to implement this. For now I have copied my fork into the project and use that.

PierfrancescoSoffritti:
I see. I think wanting to add a function to set an arbitrary playback speed is a reasonable request.

I will add the method, but it will take some time as I need to make sure the API we are introducing is clear and won't lead to confusing behavior. (And I have limited time to spend on the library)

I have opened #1081 to keep track of this, I will get to it before the end of the year.

PR #1082, opened 2023-11-26
Failed to load WebView provider: No WebView installed

billdizl:
Failed to load WebView provider: No WebView installed 
 
if the app webview not work (may webview not installed or webview update),have webview exception.
use android-youtube-player lib developer' app will be crash, but we not need this crash . so we need try catch this exception,so Put forward this pr for this lib .
thanks ,@PierfrancescoSoffritti


PierfrancescoSoffritti:
Hi, thanks for the PR.

There are too many changes in this PR, can you please clean it up to only contain the smallest change required?

PR #1085, opened 2023-11-27
Failed to load WebView provider: No WebView installed

billdizl:
Failed to load WebView provider: No WebView installed

if the app webview not work (may webview not installed or webview update),have webview exception.
use android-youtube-player lib developer' app will be crash, but we not need this crash . so we need try catch this exception,so Put forward this pr for this lib .
thanks ,@PierfrancescoSoffritti

This pr  is smallest change !   you can pr  it


billdizl:
image
image
 so i pr this

ytproofficialapp-prog:
i have solution  ytproofficialapp@gmail.com

PR #1120, opened 2024-03-16
fix(core-sample-app): gradle build

ryanleecode:
Loaded up `core-sample-app` in android studio and gradle sync didn't work until i applied these changes.


PierfrancescoSoffritti:
Sorry for getting to this so late. I don't think these changes are necessary, the module seems to build find as it is. Let me know if you think i'm missing something.

Issue #1137, opened 2024-04-25
Add a util that extract video id from YouTube url


MahmoudMabrok:
# Feature Request

### Describe your idea
a util function to extract video id from url

### Expected behavior and realization
`playerView.loadFromUrl(url, startTime)`

PierfrancescoSoffritti:
Sure, we could add that. What would be the benefit over the current functions based on id? My only concern is polluting the API too much, with nearly duplicated functions.

MahmoudMabrok:
My current app and most of apps I dealt with, we had the links of videos as full url not as id only, so every user to lib will create that util and add unit tests to it

So I thought to have it, it is like time utils

MahmoudMabrok:
Oh, sorry.

It should be a separate util not as extension to YouTube player as I shared. 

LinksUtil.getIDFromUrl.

PierfrancescoSoffritti:
Makes sense, feel free to send a pull request :) I think it should probably be in the form of new functions or extension functions.

MahmoudMabrok:
@PierfrancescoSoffritti  I did the PR and added Unit tests for it, please check it.

MahmoudMabrok:
I am closing it, as you see it is not needed.

PR #1139, opened 2024-04-26
feat: add getVideoIdFromUrl to get id

MahmoudMabrok:
apply [issues/1137](https://github.com/PierfrancescoSoffritti/android-youtube-player/issues/1137).

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
@@ -1,0 +1,25 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils
+
+/**
+ * Extracts the video ID from a YouTube URL.
+ *
+ * @param url The YouTube URL.
+ * @return The video ID, or the original URL if the video ID cannot be found.
+ */
+fun getVideoIdFromUrl(url: String): String? {
+    val startIndex = url.indexOf("?v=")
+
+    if (startIndex == -1) {
+        return null
+    }
+
+    val linkParamSymbolIndex = url.indexOf("&")
+
+    val lastIndex = if (linkParamSymbolIndex == -1) {
+        url.length
+    } else {
+        linkParamSymbolIndex
+    }
+
+    return url.substring(startIndex + 3, lastIndex)
+}
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
@@ -3,8 +3,8 @@ package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils
 /**
  * Extracts the video ID from a YouTube URL.
  *
- * @param url The YouTube URL.
- * @return The video ID, or the original URL if the video ID cannot be found.
+ * @param url The YouTube video url.
+ * @return The video ID, or null if the video ID cannot be found.
  */
 fun getVideoIdFromUrl(url: String): String? {
     val startIndex = url.indexOf("?v=")
--- a/core/src/test/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/UrlHelperTest.kt
+++ b/core/src/test/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/UrlHelperTest.kt
@@ -6,21 +6,21 @@ import org.junit.Test
 class UrlHelperTest {
 
     @Test
-    fun testGetVideoIdFromUrl_WithUrlHasIdAndUrlSeparator_ReturnsVideoId(){
+    fun testGetVideoIdFromUrl_WithUrlHasIdAndUrlSeparator_ReturnsVideoId() {
         val expected = getVideoIdFromUrl("https://www.youtube.com/watch?v=FBMAlNURLI4&t=44s")
 
         assert(expected == "FBMAlNURLI4")
     }
 
     @Test
-    fun testGetVideoIdFromUrl_WithUrlHasId_ReturnsVideoId(){
+    fun testGetVideoIdFromUrl_WithUrlHasId_ReturnsVideoId() {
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
```diff
@@ -0,0 +1,25 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils
+
+/**
+ * Extracts the video ID from a YouTube URL.
+ *
+ * @param url The YouTube URL.
+ * @return The video ID, or the original URL if the video ID cannot be found.
+ */
+fun getVideoIdFromUrl(url: String): String? {
```
PierfrancescoSoffritti, line 9: This method should probably use java.net.URL class.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
```diff
@@ -0,0 +1,25 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils
+
+/**
+ * Extracts the video ID from a YouTube URL.
+ *
+ * @param url The YouTube URL.
+ * @return The video ID, or the original URL if the video ID cannot be found.
```
PierfrancescoSoffritti, line 7: It should return null if the id can't be found.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+fun getVideoIdFromUrl(url: String): String? {
+    val startIndex = url.indexOf("?v=")
```
PierfrancescoSoffritti, line 10: I think the rest of the code is formatted using two spaces instead of tabs, please be consistent. Same for all the other files.

core/src/test/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/UrlHelperTest.kt
```diff
@@ -0,0 +1,28 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core
+
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.utils.getVideoIdFromUrl
+import org.junit.Test
+
+class UrlHelperTest {
+
+    @Test
+    fun testGetVideoIdFromUrl_WithUrlHasIdAndUrlSeparator_ReturnsVideoId(){
```
PierfrancescoSoffritti, line 9: Add spaces before curly braces.

PierfrancescoSoffritti, line 9: Remove all the methods, write a single method called `testGetVideoIdFromUrl`

MahmoudMabrok: we should have a separate test method for each scenario, or do you mean another thing ?

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
MahmoudMabrok: we normally will get data from api or any source as string url, so I think we should go with string rather than make our lib user to convert it to URL instance

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
MahmoudMabrok: I used IDE formatter, hope it be fine now, if you have a rules of formation or linting please share to apply

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
PierfrancescoSoffritti: The method should use URL internally to parse the string. Manual parse is error prone.

core/src/test/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/UrlHelperTest.kt
PierfrancescoSoffritti: We can have a single function testing multiple cases. These different functions can be merged into one, with all the tests in it.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
PierfrancescoSoffritti: I still see a 4 spaces indentation, it should be 2 spaces. I think this change is simple enough, does not require specific formatting rules.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/utils/UrlHelper.kt
MahmoudMabrok: could you share with me possible errors can be occur?
I think we do not need it.

core/src/test/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/UrlHelperTest.kt
MahmoudMabrok: yes we can have, but we should not each test case should be have a single test/assert/logic to check. 

also, what impact of having them **separated** (which make code more readable, and we can get the business from name of test cases) ?

PierfrancescoSoffritti:
Thanks for your PR. On second thought, I don't think it's this library's business to do URL parsing. Users of the library should be responsible for handling this.

We don't want to have to maintain logic that parses URLs.

PR #1140, opened 2024-04-28
chore: Delete core-sample-app/apk directory

MahmoudMabrok:
version control shouldnot have binary files, apk can be added as release or make pipeline generate it


PierfrancescoSoffritti:
Thanks for the suggestion, but it's easier to have it in version control. It allows to share links easily.

PR #1167, opened 2024-09-05
Pass videoId during initialization

talklittle:
Redo of #1166. This is an optional optimization to allow the API to load the video assets as early as possible during initialization.

Cases I tested (all working as expected):

1. Initialize with videoId, and cueVideo on ready
2. Initialize with videoId, without cueVideo
3. Initialize with null videoId, and cueVideo on ready
4. Initialize with null videoId, without cueVideo
5. Initialize with a *different* videoId, and cueVideo on ready
6. Pass the videoId from XML, without cueVideo nor `initialize()`

Case 1: Works as expected  
Case 2: Appears identical to case 1  
Case 3: Works as expected; noticed a very quick black background flicker before the correct cover image appeared  
Case 4: [As expected](https://github.com/PierfrancescoSoffritti/android-youtube-player/pull/1166#issuecomment-2330832481), shows a black screen with YouTube play button, and error message when pressing the play button (error code `INVALID_PARAMETER_IN_REQUEST`)  
Case 5: The cued video is loaded correctly. The different video used in initialization may have quickly flickered on the screen, similar to Case 3.  
Case 6: Works as expected; appears identical to cases 1 and 2

PR #1169, opened 2024-09-06
Pass videoId during initialization

talklittle:
Copy of #1167 but targeting `dev` branch.

This is an optional optimization to allow the API to load the video assets as soon as possible after initialization.


PierfrancescoSoffritti:
Thanks, sorry for making you send it twice :)

PR #1171, opened 2024-09-09
IFramePlayerOptions: Add "start", deprecate "modestBranding"

talklittle:
Add `start` to specify initial offset in seconds, similar to `seekTo`, and similar to the argument in `loadVideo` and `cueVideo`, but without having to call those methods.

Deprecate `modestBranding` — see deprecation announcement https://developers.google.com/youtube/player_parameters#release_notes_08_15_2023

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,14 @@ If set to 0: the player will show captions.
 
 If set to 1: the player won't show captions.
 
+##### `start`
+This parameter causes the player to begin playing the video at the given number of seconds from the start of the video. The parameter value is a positive integer.
+
+It works similarly to the `seekTo` method, and the `startSeconds` argument of `loadVideo` and `cueVideo`.
+
+##### `end`
+This parameter specifies the time, measured in seconds from the beginning of the video, when the player should stop playing the video. The parameter value is a positive integer.
+
 ### Full screen
 The full screen button can be added to the player by using `IFramePlayerOptions`
 
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
@@ -32,7 +32,6 @@ class IFramePlayerOptions private constructor(private val playerOptions: JSONObj
       private const val REL = "rel"
       private const val SHOW_INFO = "showinfo"
       private const val IV_LOAD_POLICY = "iv_load_policy"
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/options/IFramePlayerOptions.kt
```diff
@@ -30,13 +30,12 @@ class IFramePlayerOptions private constructor(private val playerOptions: JSONObj
...
       internal const val ORIGIN = "origin"
       private const val REL = "rel"
-      private const val SHOW_INFO = "showinfo"
       private const val IV_LOAD_POLICY = "iv_load_policy"
-      private const val MODEST_BRANDING = "modestbranding"
       private const val CC_LOAD_POLICY = "cc_load_policy"
       private const val CC_LANG_PREF = "cc_lang_pref"
       private const val LIST = "list"
       private const val LIST_TYPE = "listType"
+      private const val START = "start"
```
PierfrancescoSoffritti, line 38: Since we are adding `start`, do you mind adding `end` as well?

PierfrancescoSoffritti:
Thanks for the PR, could you update [this section in the doc](https://github.com/PierfrancescoSoffritti/android-youtube-player/tree/dev?tab=readme-ov-file#supported-options) with the new attributes? thanks!

talklittle:
Thanks for the review. Added `end` and updated readme.

README.md
```diff
@@ -360,6 +360,14 @@ If set to 0: the player will show captions.
 
 If set to 1: the player won't show captions.
 
+##### `start`
+This parameter causes the player to begin playing the video at the given number of seconds from the start of the video. The parameter value is a positive integer.
+
+It works similarly to the `seekTo` method, and the `startSeconds` argument of `loadVideo` and `cueVideo`.
```
PierfrancescoSoffritti, line 366: What happens if you define both `start` and `startSeconds`? 

Why would one prefer using one over the other?

talklittle: IFramePlayerOptions `start` is useful in conjunction with a playlist, where `loadVideo`/`cueVideo` will override the playlist behavior and the player treats it as a single video instead: The playlist navigation button disappears.

To prevent that, you can pass the video ID during initialization, and the playlist using IFramePlayerOptions, and entirely skip calling `loadVideo`/`cueVideo`.

You could achieve the same using `seekTo` but that requires an extra JS call.

PierfrancescoSoffritti: Ok, sounds good!

PierfrancescoSoffritti:
LGTM, thanks for the contributions!

PR #1209, opened 2025-03-27
Add capability to query YouTube player mute state

dachev:
- Add isMuted() method to YouTubePlayer interface
- Implement isMuted() method in YouTubePlayerImpl class
- Track mute state with private boolean variable
- Update mute() and unMute() methods to maintain state

This enhancement allows developers to check if the player is currently muted, enabling better UI feedback and toggle functionality. It's particularly useful for implementing app-level mute preferences based on user actions, so apps can remember and respect user audio preferences across video changes or app restarts.

### DIFF
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
@@ -1,0 +1,8 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks
+
+/**
+ * A callback accepting a Boolean value
+ */
+fun interface BooleanCallback {
+    fun accept(value: Boolean)
+}
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
@@ -1,4 +1,4 @@
-package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 
 /**
  * A callback accepting a Boolean value
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
@@ -3,6 +3,6 @@ package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 /**
  * A callback accepting a Boolean value
  */
-fun interface BooleanCallback {
+fun interface BooleanProvider {
     fun accept(value: Boolean)
 }
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
@@ -1,5 +1,6 @@
 package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks.BooleanCallback
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.YouTubePlayerListener
 
 /**
@@ -38,6 +39,9 @@ interface YouTubePlayer {
 
   fun mute()
   fun unMute()
+
+  /** Returns true if the player is muted, false otherwise. */
+  fun isMutedAsync(callback: BooleanCallback)
 
   /**
    * @param volumePercent Integer between 0 and 100
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
@@ -1,6 +1,5 @@
 package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 
-import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.BooleanCallback
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.YouTubePlayerListener
 
 /**
@@ -41,7 +40,7 @@ interface YouTubePlayer {
   fun unMute()
 
   /** Returns true if the player is muted, false otherwise. */
-  fun isMutedAsync(callback: BooleanCallback)
+  fun isMutedAsync(callback: BooleanProvider)
 
   /**
    * @param volumePercent Integer between 0 and 100
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
@@ -38,7 +38,6 @@ interface YouTubePlayer {
 
   fun mute()
   fun unMute()
-  fun isMuted(): Boolean
 
   /**
    * @param volumePercent Integer between 0 and 100
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
@@ -1,6 +1,6 @@
 package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 
-import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks.BooleanCallback
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.BooleanCallback
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.YouTubePlayerListener
 
 /**
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
@@ -38,6 +38,7 @@ interface YouTubePlayer {
 
   fun mute()
   fun unMute()
+  fun isMuted(): Boolean
 
   /**
    * @param volumePercent Integer between 0 and 100
--- a/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerBridge.kt
+++ b/core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerBridge.kt
@@ -5,7 +5,6 @@ import android.os.Looper
 import android.text.TextUtils
 import android.webkit.JavascriptInterface
 import androidx.annotation.RestrictTo
-import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.BooleanCallback
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.YouTubePlayerListener
 import java.util.Random
 import java.util.concurrent.ConcurrentHashMap
@@ -51,7 +50,7 @@ class YouTubePlayerBridge(private val youTubePlayerOwner: YouTubePlayerBridgeCal
   }
 
   private val mainThreadHandler: Handler = Handler(Looper.getMainLooper())
-  private val booleanCallbacks = ConcurrentHashMap<String, BooleanCallback>()
+  private val booleanCallbacks = ConcurrentHashMap<String, BooleanProvider>()
 
   interface YouTubePlayerBridgeCallbacks {
     val listeners: Collection<YouTubePlayerListener>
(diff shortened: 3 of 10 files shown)


PierfrancescoSoffritti:
Hi, thanks for your PR.

This approach will lead to issues, since it requires duplicating the state. The source of truth is the IFrame Player, storing the boolean on the JVM side can end up with the two going out of sync.

The correct way of doing this would be to add a way to query the IFrame Player.

dachev:
Good catch! This is what happens when you code around midnight. I'll update it to directly query the WebView instead of tracking state locally. Thanks for the review.

dachev:
@PierfrancescoSoffritti It'll be hard to add this to the `YouTubePlayer` interface because:

1. the `ChromecastYouTubePlayer` implementation (and testing) would take some time and effort
2. a default implementation wouldn't make sense for a method like `isMuted`

My solution was to implement a second interface (`YouTubePlayerStateProvider`) which will allow people to typecast. Let me know if you think there is a better way to do this.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerStateProvider.kt
```diff
@@ -0,0 +1,9 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
+
+interface YouTubePlayerStateProvider {
```
PierfrancescoSoffritti, line 3: Instead of adding a separate interface:

1. Add this to YouTubePlayer

```kotlin
/** Returns true if the player is muted, false otherwise. */
  fun isMutedAsync(callback: BooleanCallback)
```

2. Add a new file called YouTubePlayerExt

```kotlin
/** Returns true if the player is muted, false otherwise. */
suspend fun YouTubePlayer.isMuted(): Boolean = suspendCoroutine { continuation ->
  isMutedAsync { isMuted ->
    continuation.resume(isMuted)
  }
}
```

3. Unfortunately because of java7 add a new file called Callbacks with

```
/**
 * A callback accepting a Boolean value.
 *
 * This functional interface is only required to support Java 7 and below.
 */
fun interface BooleanCallback {
  fun accept(value: Boolean)
}
```

4. in ChromecastYouTubePlayer it's ok to not implement the method for now

```kotlin
override fun isMutedAsync(callback: BooleanCallback) {
    throw NotImplementedError("isMutedAsync is not implemented in ChromecastYouTubePlayer")
  }
```

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
```diff
@@ -66,6 +83,46 @@ private class YouTubePlayerImpl(private val webView: WebView) : YouTubePlayer {
...
     mainThread.post { loadUrl("javascript:$function(${stringArgs.joinToString(",")})") }
   }
+
+  /**
+   * Executes JavaScript and gets the return value via a callback.
+   *
+   * @param javascript The JavaScript code to execute
+   * @param callback Function that will be called with the result
+   */
+  private fun WebView.evaluateWithResult(javascript: String, callback: (String) -> Unit) {
```
PierfrancescoSoffritti, line 93: There are a few issues with this:

1. 
`evaluateJavascript` seem to cause the player to glitch and skip some frames.

2.
`addJavascriptInterface` doesn't work. The page needs to be re-loaded after this is added.

I think unfortunately a different approach might be required.

1. We should leverage the existing `YouTubePlayerBridge` to receive data from JS.
2. Data intended as return value for a method invocation like `isMute` should be place into a queue.
3. When methods like `isMute` are invoked, they should poll the queue until the data they're looking for becomes available, or timeout.

PierfrancescoSoffritti:
Hi, i have made some changes locally to try out your PR (like adding a sample app). If you allow edits from maintainers I can push them directly to your PR.

As mentioned in the comments, i'm afraid this approach won't work. A more convoluted solution is required.

PierfrancescoSoffritti (commented):
Hi, i have made some changes locally to try out your PR (like adding a sample app). If you allow edits from maintainers I can push them directly to your PR.

As mentioned in the comments, i'm afraid this approach won't work. A more convoluted solution is required.

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
```diff
@@ -0,0 +1,8 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks
+
+/**
+ * A callback accepting a Boolean value
+ */
+fun interface BooleanCallback {
```
PierfrancescoSoffritti, line 6: change to

```kotlin
/**
 * A callback that accepts a Boolean value.
 *
 * This interface is only required to support Java 7 and below.
 */
fun interface BooleanProvider {
    fun accept(value: Boolean)
}
```

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayer.kt
```diff
@@ -1,5 +1,6 @@
 package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
 
+import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks.BooleanCallback
```
PierfrancescoSoffritti, line 3: Remove import

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerBridge.kt
```diff
@@ -152,6 +156,22 @@ class YouTubePlayerBridge(private val youTubePlayerOwner: YouTubePlayerBridgeCal
     youTubePlayerOwner.listeners.forEach { it.onVideoId(youTubePlayerOwner.getInstance(), videoId) }
   }
 
+  @JavascriptInterface
```
PierfrancescoSoffritti, line 159: Let's move all the new logic to a new class:

```kotlin
/** Bridge used to extract values from Javascript and pass them to the YouTubePlayer. */
internal class YouTubePlayerCallbacks {
  private val mainThreadHandler: Handler = Handler(Looper.getMainLooper())

  /** Callbacks registered by clients of this class to retrieve boolean values form Javascript. */
  private val booleanCallbacks = ConcurrentHashMap<Long, BooleanProvider>()

  private val requestId = AtomicLong(0)

  /**
   * Registers a callback to be called when a boolean value is received from Javascript.
   * @return the requestId for this callback.
   */
  fun registerBooleanCallback(callback: BooleanProvider): Long {
    val requestId = requestId.incrementAndGet()
    booleanCallbacks[requestId] = callback
    return requestId
  }

  @JavascriptInterface
  fun sendBooleanValue(requestId: Long, value: Boolean) {
    mainThreadHandler.post {
      val callback = booleanCallbacks.remove(requestId)
      callback?.accept(value)
    }
  }
}
```

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerExt.kt
```diff
@@ -0,0 +1,11 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player
+
+import kotlin.coroutines.resume
+import kotlin.coroutines.suspendCoroutine
+
+/** Returns true if the player is muted, false otherwise. */
+suspend fun YouTubePlayer.isMuted(): Boolean = suspendCoroutine { continuation ->
```
PierfrancescoSoffritti, line 7: Use one line

```kotlin
/** Returns true if the player is muted, false otherwise. */
suspend fun YouTubePlayer.isMuted(): Boolean = suspendCoroutine { continuation ->
    isMutedAsync { isMuted -> continuation.resume(isMuted) }
}
```

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
```diff
@@ -40,6 +41,12 @@ private class YouTubePlayerImpl(private val webView: WebView) : YouTubePlayer {
   override fun setShuffle(shuffle: Boolean) = webView.invoke("setShuffle", shuffle)
   override fun mute() = webView.invoke("mute")
   override fun unMute() = webView.invoke("unMute")
+  override fun isMutedAsync(callback: BooleanCallback) {
+    val bridge = (webView as WebViewYouTubePlayer).getBridge()
```
PierfrancescoSoffritti, line 45: Remove this getter can cast. Instead pass `YouTubePlayerCallbacks` to the constructor of `YouTubePlayerImpl`.

```kotlin
private class YouTubePlayerImpl(
  private val webView: WebView,
  private val callbacks: YouTubePlayerCallbacks
) : YouTubePlayer {
}
```

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
```diff
@@ -40,6 +41,12 @@ private class YouTubePlayerImpl(private val webView: WebView) : YouTubePlayer {
...
+    val requestId = bridge.registerBooleanCallback(callback)
+
```
PierfrancescoSoffritti, line 47: Remove empty space

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
```diff
@@ -153,6 +162,8 @@ internal class WebViewYouTubePlayer constructor(
 
     super.onWindowVisibilityChanged(visibility)
   }
+
+  internal fun getBridge(): YouTubePlayerBridge = youTubePlayerBridge
```
PierfrancescoSoffritti, line 166: Remove getter

core/src/main/res/raw/ayp_youtube_player.html
```diff
@@ -186,6 +186,10 @@
     function setShuffle(shuffle) {
       player.setShuffle(shuffle);
     }
+    function getMuteState(requestId) {
```
PierfrancescoSoffritti, line 189: Add empty space between `setShuffle` and `getMuteValue`

PierfrancescoSoffritti, line 189: Rename to `getMuteValue`

PierfrancescoSoffritti (commented):
Looks very nice, thanks! Just a few minor changes

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/Callbacks.kt
```diff
@@ -0,0 +1,8 @@
+package com.pierfrancescosoffritti.androidyoutubeplayer.core.player.callbacks
```
PierfrancescoSoffritti, line 1: Fix package name

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerBridge.kt
```diff
@@ -48,6 +50,7 @@ class YouTubePlayerBridge(private val youTubePlayerOwner: YouTubePlayerBridgeCal
   }
 
   private val mainThreadHandler: Handler = Handler(Looper.getMainLooper())
+  private val booleanCallbacks = ConcurrentHashMap<String, BooleanProvider>()
```
PierfrancescoSoffritti, line 53: Remove

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/YouTubePlayerBridge.kt
```diff
@@ -6,6 +6,8 @@ import android.text.TextUtils
 import android.webkit.JavascriptInterface
 import androidx.annotation.RestrictTo
 import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.YouTubePlayerListener
+import java.util.Random
+import java.util.concurrent.ConcurrentHashMap
```
PierfrancescoSoffritti, line 10: Remove

core/src/main/java/com/pierfrancescosoffritti/androidyoutubeplayer/core/player/views/WebViewYouTubePlayer.kt
```diff
@@ -25,7 +27,10 @@ import java.io.InputStreamReader
 import java.util.*
 
 
-private class YouTubePlayerImpl(private val webView: WebView) : YouTubePlayer {
+private class YouTubePlayerImpl(
+        private val webView: WebView,
```
PierfrancescoSoffritti, line 31: Fix indentation. Use two spaces.

PierfrancescoSoffritti (commented):
Last few nits, then we can merge :)

PierfrancescoSoffritti:
You're right. How about we just call it `Providers`?

PierfrancescoSoffritti (approved):
LGTM, thanks for doing this!

dachev:
The pleasure was all mine. Thanks!

dachev:
@PierfrancescoSoffritti BTW, I ran into some multidex-ing issues in the example apps while I was trying to build the codebase for testing purposes. Do you want me to send you a PR for these?

PierfrancescoSoffritti:
I noticed you had that. I am not encountering that problem. I have updated agp to the latest version on the dev branch, try pulling and see if that solves the problem.

PierfrancescoSoffritti:
I see, what if you try to deploy to a newer API? I suspect this is happening because of API 29.

PierfrancescoSoffritti:
Uhm then i'm not sure why you're getting that. Either way seems strange, the sample app should not exceed the max method count. If it does we should figure out why.

dachev:
@PierfrancescoSoffritti It liiks like it's the Chromecast and the Material dependencies.

Screenshot 2025-04-02 at 8 02 14 pm

Issue #1222, opened 2025-06-11
How to implement with Jetpack Compose


tiendz2003:
I get some trouble when implement the library when use jetpack compose.Did anyone use jetpack compose can give me some examples to explore.I really need this.

PierfrancescoSoffritti:
It would also be nice to add an example in the sample app. #1007

barnhill:
Im experiencing the same issue with 12.1.1 and getting this error in the logs that doesnt show up on 12.1.0

```
[INFO:CONSOLE:196] "Uncaught Error: Invalid video id", source: https://www.youtube.com/s/player/14397202/www-widgetapi.vflset/www-widgetapi.js (196)
```

Any time I see this the player will not play.

barnhill:
Will try this later today and confirm

anilInvestKraft:
is anybody face issue that all the control icons shift to the left sid of screen.

barnhill:
No but with my compose wrapper setting the controls to hide still allows controls to show.

Issue #1007, opened 2023-06-15
Using it in Jetpack Compose with full screen capabilities


aad4:
# Question / Problem

### Summary
I try to use the library in a Jetpack Compose application for the following use cases:
1. Multiple video players can be displayed on a single screen, adhering to the spacing between videos that I provide
1. When not all videos fit on the screen, vertical scrolling is used to scroll through all videos of the page
1. The app allows for rotating the screen (from portrait to landscape and vice versa)
1. A video player allows for switching to full screen and back again
1. When switching to full screen, the sound/video continues to play (almost) without interruption. Same when switching back to normal view.

Questions I have:
1. Are the use cases mentioned above even possible with the library?
1. How to get full screen mode working in Jetpack Compose?
1. Why is the landscape mode resulting in strange layout/behavior (see below)?

### What I've tried

I created a [repository with an Jetpack Compose example app](https://github.com/aad4/androidyoutubeplayer-in-jetpack) so you can see what I tried.

The first commit is a simple version without full screen functionality; it renders just 1 video player on the screen. When rotating to landscape, the video player is not fully shown; the bottom part with the controls is (mostly) missing, and I also cannot scroll to the bottom to make them visible.

The second commit extended the app by showing 2 video players with some spacing in between, and allows for full screen mode. When clicking on the full screen button of a video player, it hides the (small) player as expected but does not show the full screen player, and the sound stops. Another issue is that in landscape mode the spacing between the two videos is gone, both videos are only partially visible, and scrolling is again not possible.

PierfrancescoSoffritti:
> Are the use cases mentioned above even possible with the library?

Yes

> How to get full screen mode working in Jetpack Compose?

You take the `fullScreenView` provided by `FullscreenListener#onEnterFullscreen` and add it as a view that covers the entire screen.

> When rotating to landscape, the video player is not fully shown; the bottom part with the controls is (mostly) missing, and I also cannot scroll to the bottom to make them visible.

It seems like you're not setting the view containing the video to fill the screen?

EncMukesh:
I get a white screen without video when changing the Orientation but the audio is playing.

rafaelekol:
@PierfrancescoSoffritti 
It would great top have samples with Jetpack Compose.
I also got confused how to get fullscreen mode using Compose.

justinmulligan:
Is there a Jetpack Compose example yet? One that includes full screen as well?

[...]

PR #1270, opened 2025-10-24
Add Compose example #1222 #1007

robertlinsener:


PierfrancescoSoffritti:
Thanks a lot for the PR. I have made a small change to add the ability to play different videos, so it's clear the same instance of `YouTubePlayer` needs to be reused.

Issue #1115, opened 2024-03-13
AUDIO_PERFORMANCE_MODE_LOW_LATENCY


mariam-wahba:
# Bug Report

### Description of the Bug
Hello, I'm facing a problem. I'm using     implementation 'com.pierfrancescosoffritti.androidyoutubeplayer:core:12.1.0' and implemented in my android app project. It works so well if i ran on virtual device android 11 while on physical android 11 device it is like continuous cutting while playing the video and i find this error in my console 15:12:18.255 10720-10822 AudioStreamBuilder com.xlab.andsign D build() MMAP not used because AAUDIO_PERFORMANCE_MODE_LOW_LATENCY not requested.I don't know why this happens. Can anyone help me?

### Environment details
Android 11
Api level: 26 to 34

Tested devices:

Youtube Player Library Version:   implementation 'com.pierfrancescosoffritti.androidyoutubeplayer:core:12.1.0'

### Steps to reproduce the bug
import static androidx.constraintlayout.helper.widget.MotionEffect.TAG;

import android.content.Context;
import android.util.Log;
import android.view.View;
import android.webkit.WebView;
import android.widget.FrameLayout;
import android.widget.Toast;

import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.PlayerConstants;
import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.YouTubePlayer;
import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.listeners.AbstractYouTubePlayerListener;
import com.pierfrancescosoffritti.androidyoutubeplayer.core.player.views.YouTubePlayerView;
import com.xlab.Entities.PlayListRecord;

import androidx.annotation.NonNull;

public class YoutubeView extends FrameLayout {
	private PlayListRecord playListRecord;
	private final static int RETRY_DELAY = 10000;
	private YoutubeErrorListener errorListener;
	private boolean isError;
	private boolean isInitialized;
	private boolean isNetError;
	private boolean isPlayStarted;
	private boolean isStopped;
	private boolean isTerminated;
	boolean isPlayRetryQueued;
	private final boolean isTraceEnabled;
	private long lastPlayTime;
	private String videoId;
	private WebView webView;
	private YouTubePlayerView youTubePlayerView;

	public YoutubeView(Context context, PlayListRecord playListRecord) {
		super(context);
		isTraceEnabled = false;
		isInitialized = false;
		isNetError = false;
[...]

PR #1309, opened 2026-05-02
Expose software/hardware layer toggle on YouTubePlayerView

### LINKED ISSUE #1121: are there any specific configurations or recommendations for dealing with audio playback on Android 11 giada jhs 558 and ensuring low-latency performance mode is enabled?
# Question / Problem

jim-daf:
### Issue

Closes #1121.

The reporter sees continuous video and audio cutting on a giada DN74 (Android
11) tablet while the same code runs fine on the Android 11 emulator and on a
Realme 5 Pro. The logcat is dominated by:

```
AudioStreamBuilder build() MMAP not used because AAUDIO_PERFORMANCE_MODE_LOW_LATENCY not requested.
AudioStreamLegacy  onAudioDeviceUpdate() request DISCONNECT in data callback, device 9 => 0
AudioStreamLegacy  processCallbackCommon() data, stream disconnected
AudioTrack         processAudioBuffer(976): EVENT_MORE_DATA requested 4488 bytes but callback returned -1 bytes
```

That sequence is the documented Android 11 WebView / GPU-driver interaction
where a frame produced through the hardware accelerated WebView pipeline
tears down the AAudio output stream on devices whose low-latency audio path
is incomplete (a number of low-end Rockchip / MTK media boxes are affected,
including the giada DN74). Issue #1115 reports the same `AAUDIO_PERFORMANCE_MODE_LOW_LATENCY`
log signature.

### Why a library change

The Android-recommended workaround is `View.setLayerType(LAYER_TYPE_SOFTWARE, null)`
on the offending WebView, which bypasses the GPU compositor for that view
and keeps audio playback stable. The wider workaround "disable hardware
acceleration on the whole Activity" is already mentioned in the README under
**Useful info > Hardware acceleration**, but it has the well-known side
effect of producing a black image while audio plays normally, so it is not a
real fix.

The targeted, per-view workaround was unreachable from application code
because `WebViewYouTubePlayer` is `internal`. This PR exposes it.

### Change

`YouTubePlayerView` gains two opt-in methods that delegate to
`View.setLayerType` on the embedded WebView:

```kotlin
fun setWebViewSoftwareLayerType()
fun setWebViewHardwareLayerType()
```

Typical usage:

```kotlin
val youTubePlayerView: YouTubePlayerView = findViewById(R.id.youtube_player_view)
if (Build.MODEL.equals("DN74", ignoreCase = true)) {
    youTubePlayerView.setWebViewSoftwareLayerType()
}
```

The README "Hardware acceleration" section is updated with a short note that
links the new hook to playback glitches on Android 11 devices with broken
AAudio paths.

### Compatibility

* Purely additive API. No existing public symbol is modified or removed.
* Default behaviour (hardware-accelerated layer) is unchanged.
* The new methods just forward to the standard Android `View.setLayerType`
  API, so they cannot regress existing tests or sample apps.

Marking this as a draft so you can decide whether you would prefer the
methods to live on `LegacyYouTubePlayerView` only or be exposed as an XML
attribute as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -821,6 +821,8 @@ If you need to disable hardware acceleration in your application, you can enable
 
 Disabling hardware acceleration on the Activity containing `YouTubePlayerView` may result in some weird behavior. The one I have observed so far shows a black image in the player, while the audio is playing normally.
 
+On a small number of devices (notably some low-end Android 11 media boards whose AAudio low-latency path is incomplete) the hardware accelerated WebView pipeline tears down the audio output stream during playback, producing continuous video/audio cutting. As a per-view workaround you can switch the embedded WebView to a software-rendered layer without touching the rest of the Activity by calling `YouTubePlayerView.setWebViewSoftwareLayerType()`. Use `YouTubePlayerView.setWebViewHardwareLayerType()` to revert.
+
 ### Play YouTube videos in the background
 With this library it's easy to play YouTube videos when the app is not visible. In order to do that you simply have to not call `youTubePlayer.pause()` when the Activity is being paused or stopped and enable background playback by calling `YouTubePlayerView.enableBackgroundPlayback(true)`.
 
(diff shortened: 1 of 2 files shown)


PierfrancescoSoffritti:
Why is this change needed?

